PR #69, opened 2020-07-01
first draft of the trading contract

joshuahannan:
We've been seeing a lot of requests in the discord channels for automated trading so that you don't have to trust the player who you are trading with. This allows a user to create a trading center in their account where they can put a moment up for trade and specify what kind of moment they are willing to trade it for, using id, setID, playID, and serial number. 

When someone sends a moment that meets the criteria that the originator specified, both moments are deposited into the participants' collections


tpetrychyn:
closing as stale

PR #84, opened 2020-08-03
Changes market to use capabilities

joshuahannan:
I have modified the Top Shot market contract to use a capability for the owner's collection instead of storing the actual moments. This solves a few problems.
* Apps that listen for withdraw events will not think a moment has left a user's collection if they put it on sale.
* Moments can still be interacted with in many different ways through the standard collection while they are on sale

The new market contract is passing all of the tests.

This PR also creates a new set of transactions that are used for the new version of the contract.
They all have a `V2` in the names of the functions that are used to generate them.

Using this new version of the contract will not be able to be done with a direct upgrade to the existing smart contract. It will need to be deployed separately and the existing data from the old contract will either need to be managed in parallel or migrated to the new version.


rrrkren:
Are going to be able to directly upgrade to this new contract? Or do we need to do a migration/reset of the p2p marketplace?

joshuahannan:
Using this new version of the contract will not be able to be done with a direct upgrade to the existing smart contract. It will need to be deployed separately and the existing data from the old contract will either need to be managed in parallel or migrated to the new version.

lib/go/templates/marketV2_templates.go
```diff
@@ -0,0 +1,159 @@
...
+		transaction {
+			prepare(acct: AuthAccount) {
+				let ownerCapability = acct.getCapability<&{FungibleToken.Receiver}>(/public/%[3]sReceiver)!
+				let beneficiaryCapability = getAccount(0x%[2]s).getCapability<&{FungibleToken.Receiver}>(/public/%[3]sReceiver)!
+
+				let ownerCollection: Capability<&TopShot.Collection> = acct.link<&TopShot.Collection>(/private/MomentCollection, target: /storage/MomentCollection)!
+
+				let collection <- Market.createSaleCollection(ownerCollection: ownerCollection, ownerCapability: ownerCapability, beneficiaryCapability: beneficiaryCapability, cutPercentage: %[4]f)
+				
+				acct.save(<-collection, to: /storage/topshotSaleCollection)
```
rrrkren, line 31: shall we append a version to the storage path if we want to have both old version and new version functional at the same time? It would help managing the migration a lot too. cc @EricLin2004

rrrkren:
@joshuahannan can you make the transaction compatible with both contracts?

joshuahannan:
@rrrkren I can make some of them compatible with both, but there are a few that have to be different

Can be compatible with both: 

GenerateChangePercentageScript
GenerateChangeOwnerReceiverScript
GenerateBuySaleScript
GenerateMintTokensAndBuyScript
GenerateInspectSaleScript
GenerateInspectSalePercentageScript
GenerateInspectSaleLenScript
GenerateInspectSaleMomentDataScript

Cannot be compatible with both:

GenerateCreateSaleScript
GenerateStartSaleScript
GenerateCreateAndStartSaleScript
GenerateCancelSaleScript
GenerateChangePriceScript

What do you think?

rrrkren:
start sale I think is okay. why can't we do CancelSale? can we fetch it from the old collection if it exists first and then go to the new collection if not then panic?

joshuahannan:
@rrrkren That would be a new transaction. I can make that if you want!

rrrkren:
yeah we are ok with sending different transactions

joshuahannan:
@rrrkren I added the template for upgrading an old sale listing to a new one. I don't have time to test it because I'm really busy with flow token delivery and staking stuff, but it is there if you want to try it out. `GenerateUpgradeSaleV2Script` and `transactions/market/upgrade_sale.cdc`

README.md
```diff
@@ -296,12 +308,30 @@ and if they sent the correct amount, they get the Moment back.
    Emitted when a seller changes the percentage cut that is taken
    from their sales and sent to a beneficiary.
 
+### Different Versions of the Market Contract
+
+There are two versions of the Top Shot Market Contract.
+`MarketTopShotv1.cdc` is the original version of the contract that was used
+for the first set of sales in the p2p marketplace, but we made improvements
+to it which are now in `MarketTopShot.cdc`.
```
DavidRS91, line 316: I think this needs to be updated.
```suggestion
There are two versions of the Top Shot Market Contract.
`MarketTopShot.cdc` is the original version of the contract that was used
for the first set of sales in the p2p marketplace, but we made improvements
to it which are now in `MarketTopShotV2.cdc`.
```

contracts/TopShotMarketV2.cdc
```diff
@@ -0,0 +1,295 @@
+/*
+
+    TopShotMarketV2.cdc
```
DavidRS91, line 3: It seems there is a naming convention inconsistency between `MarketTopShot.cdc` and `TopShotMarket.cdc` - should we rename this `MarketTopShotV2.cdc`?

README.md
```diff
@@ -296,12 +308,30 @@ and if they sent the correct amount, they get the Moment back.
...
+put up for sale. In this version, when a moment is purchased, the sale collection
+uses the capability to withdraw the moment from the main collection and 
+returns it to the buyer.
 
 ## License 
 
 The works in these folders 
 /dapperlabs/nba-smart-contracts/blob/master/contracts/TopShot.cdc 
 /dapperlabs/nba-smart-contracts/blob/master/contracts/MarketTopShot.cdc 
+/dapperlabs/nba-smart-contracts/blob/master/contracts/MarketTopShotv1.cdc
```
DavidRS91, line 334: outdated

lib/go/templates/marketV2_templates.go
```diff
@@ -0,0 +1,387 @@
...
+
+				topshotSaleCollection.listForSale(tokenID: %[3]d, price: %[4]d.0)
+			}
+		}`
+	return []byte(fmt.Sprintf(template, topshotAddr, marketAddr, id, price))
+}
+
+// GenerateCreateAndStartSaleV2Script creates a cadence transaction that creates a Sale collection
+// and stores in in the callers account, and also puts an NFT up for sale in it
+func GenerateCreateAndStartSaleV2Script(ftAddr, topshotAddr, marketAddr, beneficiaryAddr flow.Address, tokenStorageName string, cutPercentage, price float64, tokenID int) []byte {
```
DavidRS91, line 60: It looks like we're using price as an int in v1 and then converting it to UFix64 in the template, is there a reason it switched to a float64 in v2?
https://github.com/dapperlabs/nba-smart-contracts/blob/e6cd787cebb1eb50d04dccdc27844d3bff35511b/lib/go/templates/market_templates.go#L56

joshuahannan: It might have been what I copied vs what you actually used in your app. Fee free to change it to whatever you want it to be @DavidRS91

joshuahannan:
@EricLin2004 Can I just get an approval and merge this PR? It has been in limbo for like 6 months now and I'd like to just get it merged so I don't have to keep updating it. We can just decide to use the new contract whenever we are ready.

PR #94, opened 2021-02-04
cleaning up tests and updating versions

joshuahannan:
Utilized utility functions to the tests for:

* Constructing a transaction
* Executing a Script
* Creating a Cadence UFix64

Updated to the latest Cadence versions
Added more to the makefiles


rrrkren:
can you give a summary of where flow-go-sdk is used in this repo? is it only for tests?

joshuahannan:
@rrrkren yeah it is only for tests
https://github.com/dapperlabs/nba-smart-contracts/blob/master/lib/go/test/topshot_test.go#L14

joshuahannan:
I'm actually planning on doing that in this other PR: https://github.com/dapperlabs/nba-smart-contracts/pull/96
I'd like to get this one merged first though

joshuahannan:
I parameterized all the transactions in that PR so that we can keep the sdk stuff only as part of the tests and have the templates just for getting the transaction scripts

lib/go/test/examples.go
```diff
@@ -37,69 +38,72 @@ func DownloadFile(url string) ([]byte, error) {
...
-func SignAndSubmit(
+// signAndSubmit signs a transaction with an array of signers and adds their signatures to the transaction
+// before submitting it to the emulator.
+//
+// If the private keys do not match up with the addresses, the transaction will not succeed.
+//
+// The shouldRevert parameter indicates whether the transaction should fail or not.
+//
+// This function asserts the correct result and commits the block if it passed.
+func signAndSubmit(
```
psiemens, line 81: It'd be great to get these utilities into the `flow-go-sdk/test` package: https://github.com/onflow/flow-go-sdk/tree/master/test

Or maybe the emulator?

psiemens (approved):
LGTM, nice!

lib/go/test/examples.go
joshuahannan: Yeah that is a good idea. What would be the best way to do that? Can we just put them in as is, or should we have a discussion about which parts we want to include?

PR #95, opened 2021-02-05
fix typo in args of GenerateBatchTransferMomentScript

ramtinms:
seems to be a small typo here, which causes the script to fail with 
```
expected `tokens`, got `token`
``` 
error


joshuahannan:
I'm going to close this because I fixed this along with refactoring all the other transactions in: https://github.com/dapperlabs/nba-smart-contracts/pull/96

PR #98, opened 2021-02-23
Update variable name in topshot_admin_templates.go

marcusmolchany:
i wouldn't call this a value add pr, but it does fix a typo.


joshuahannan:
Thank you for fixing this! We really appreciate it. 😃

joshuahannan (approved):

Issue #75, opened 2020-07-15
TXS: Improve transactions so that they all use correct `prepare` and `execute` syntax


joshuahannan:
### Issue To Be Solved
It is a best practice in cadence to only perform actions that are associated with the transaction signer's account in the `prepare` phase of a transaction, and perform all external interaction in the `execute` phase.

### Suggest A Solution
* Update the sample transactions in `transactions/` to use both phases.

Example: `transactions/admin/add_plays_to_set.cdc` would be changed from:
```
transaction(setID: UInt32, plays: [UInt32]) {

    prepare(acct: AuthAccount) {

        // borrow a reference to the Admin resource in storage
        let admin = acct.borrow<&TopShot.Admin>(from: /storage/TopShotAdmin)
            ?? panic("Cannot borrow a reference to the admin resource")

        // borrow a reference to the set to be added to
        let setRef = admin.borrowSet(setID: setID)

        // Add the specified play IDs
        setRef.addPlays(playIDs: plays)
    }
}
```
to:
```
import TopShot from 0xTOPSHOTADDRESS

// This transaction adds multiple plays to a set
		
transaction(setID: UInt32, plays: [UInt32]) {

    // Local variable for the topshot Admin object
    let adminRef: &TopShot.Admin

    prepare(acct: AuthAccount) {

        // borrow a reference to the Admin resource in storage
        self.adminRef = acct.borrow<&TopShot.Admin>(from: /storage/TopShotAdmin)!
    }

    execute() {

        // borrow a reference to the set to be added to
        let setRef = self.adminRef.borrowSet(setID: setID)

        // Add the specified play IDs
        setRef.addPlays(playIDs: plays)
    }
}
```

PR #103, opened 2021-03-08
Fixes issue #75 : changed transactions to use prepare/execute

timongll:
Fixes issue #75: Changed a few simple transactions from nba-smart-contracts to use prepare/execute instead of prepare or none at all


joshuahannan:
@timongll Thank you so much for working on these! Can you run the tests with `make test` also to make sure they pass?

PR #104, opened 2021-03-09
Fix Damian Lillard typo in `destroy` function's documentation

danielhour:
Referencing the NBA's player page:
https://www.nba.com/player/203081/damian-lillard


joshuahannan (approved):
Good catch! Thanks for fixing this. :)

PR #106, opened 2021-03-12
Fixed ci

timongll:


joshuahannan:
It looks like this has the same changes as #105 . Are there any differences?

timongll:
its the same as #105 but 'make ci` failed. 

in #106 did 
```
git add . 
make ci
```

3 times and then make ci passed

joshuahannan:
We can probably close this then. I'm going to merge #84 soon and that has your changes and the ci passing so this one probably isn't needed

PR #118, opened 2021-06-11
Adds V3 Market Contract and transactions

joshuahannan:
Adds TopShotMarketV3, which is similar to V2 in that it uses a capability to the user's moment collection, but it also imports the V1 market contract and stores a capability to the V1 sale collection so that a single call to the SaleCollection will check both the V1 SaleCollection and V3 SaleCollection

Adds tests for V3


README.md
```diff
@@ -363,7 +363,7 @@ and if they sent the correct amount, they get the Moment back.
 There are two versions of the Top Shot Market Contract.
```
sadief, line 363: Should we make reference to there being the intermediary `TopShotMarketV2` version - and why V3 was created on top of it?

contracts/TopShotMarketV3.cdc
```diff
@@ -0,0 +1,341 @@
...
+                if let v1Market = self.marketV1Capability {
+                    let v1MarketRef = v1Market.borrow()!
+
+                    return <-v1MarketRef.purchase(tokenID: tokenID, buyTokens: <-buyTokens)
+                } else {
+                    panic("No token matching this ID for sale!")
+                }
+            }
+            
+            destroy buyTokens // THIS LINE NEEDS TO BE REMOVED ONCE THE BUG IN CADENCE IS FIXED
```
sadief, line 235: Just for context - what is the bug in Cadence?

transactions/marketV3/purchase_both_markets.cdc
```diff
@@ -0,0 +1,56 @@
...
+        } else if let marketV1CollectionRef = seller.getCapability(/public/topshotSaleCollection)
+            .borrow<&{Market.SalePublic}>() {
+            // purchase the moment
+            let purchasedToken <- marketV1CollectionRef.purchase(tokenID: momentID, buyTokens: <-self.purchaseTokens)
+
+            // deposit the purchased moment into the signer's collection
+            receiverRef.deposit(token: <-purchasedToken)
+
+        } else {
+            destroy self.purchaseTokens // REMOVE THIS
```
sadief, line 51: Is this related to the Cadence bug as well?

contracts/TopShotMarketV3.cdc
joshuahannan: https://github.com/onflow/cadence/pull/1000
It'll be released in the next version of Cadence, and I'll update this after that is released

transactions/marketV3/purchase_both_markets.cdc
joshuahannan: Yes, it is the same bug

lib/go/templates/marketV3_templates.go
```diff
@@ -5,142 +5,144 @@ import (
...
-func GenerateCreateAndStartSaleV2Script(env Environment) []byte {
-	code := assets.MustAssetString(transactionsPath + createAndStartSaleV2Filename)
+func GenerateCreateAndStartSaleV3Script(env Environment) []byte {
+	code := assets.MustAssetString(transactionsPath + createAndStartSaleV3Filename)
 
 	return []byte(replaceAddresses(code, env))
 }
 
-// GenerateCancelSaleV2Script creates a cadence transaction that starts a sale by depositing
+// GenerateCancelSaleV3Script creates a cadence transaction that starts a sale by depositing
```
sadief, line 56: Just to double check - we would use this now over the `GenerateWithdrawFromSaleScript`  because we no longer need a a withdraw in order to delist a moment, right? (Do we need to update the comment for this one?)

README.md
joshuahannan: yeah I can add a line

lib/go/templates/marketV3_templates.go
joshuahannan: yep, I'll change the comment

contracts/TopShotMarketV3.cdc
```diff
@@ -0,0 +1,346 @@
...
+                    .deposit(from: <-beneficiaryCut)
+                
+                // Deposit the remaining tokens into the owners vault
+                self.ownerCapability.borrow()!
+                    .deposit(from: <-buyTokens)
+
+                emit MomentPurchased(id: tokenID, price: price, seller: self.owner?.address)
+
+                // Return the purchased token
+                let boughtMoment <- self.ownerCollection.borrow()!.withdraw(withdrawID: tokenID) as! @TopShot.NFT
```
DavidRS91, line 226: Will there be any logic to cancel existing sales when a moment is transferred? Based on my novice understanding of cadence, I guess a sale would exist after transfer, and fail here when someone tries to make a purchase on a transferred moment?

lib/go/contracts/contracts.go
```diff
@@ -65,14 +67,15 @@ func GenerateTopShotMarketContract(ftAddr, nftAddr, topshotAddr string) []byte {
...
+// GenerateTopShotMarketV3Contract returns a copy
+// of the third version TopShotMarketContract with the import addresses updated
+func GenerateTopShotMarketV3Contract(ftAddr, nftAddr, topshotAddr, marketAddr string) []byte {
 
-	marketCode := assets.MustAssetString(marketV2File)
+	marketCode := assets.MustAssetString(marketV3File)
 	codeWithNFTAddr := strings.ReplaceAll(marketCode, defaultNonFungibleTokenAddress, nftAddr)
 	codeWithTopshotAddr := strings.ReplaceAll(codeWithNFTAddr, defaultTopshotAddress, topshotAddr)
 	codeWithFTAddr := strings.ReplaceAll(codeWithTopshotAddr, defaultFungibleTokenAddress, ftAddr)
+	codeWithMarketV2Addr := strings.ReplaceAll(codeWithFTAddr, defaultMarketAddress, marketAddr)
```
DavidRS91, line 78:
```suggestion
	codeWithMarketV3Addr := strings.ReplaceAll(codeWithFTAddr, defaultMarketAddress, marketAddr)
```

lib/go/contracts/contracts.go
```diff
@@ -65,14 +67,15 @@ func GenerateTopShotMarketContract(ftAddr, nftAddr, topshotAddr string) []byte {
...
 
-	return []byte(codeWithFTAddr)
+	return []byte(codeWithMarketV2Addr)
```
DavidRS91, line 80:
```suggestion
	return []byte(codeWithMarketV3Addr)
```

lib/go/templates/template_utils.go
```diff
@@ -3,6 +3,8 @@ package templates
 import (
 	"fmt"
 	"strings"
+
+	_ "github.com/kevinburke/go-bindata"
```
DavidRS91, line 7: Is this import needed?

lib/go/test/market_test.go
```diff
@@ -871,24 +879,261 @@ func TestV2Market(t *testing.T) {
...
+		_ = tx.AddArgument(cadence.NewUInt64(3))
+		_ = tx.AddArgument(CadenceUFix64("50.0"))
+
+		signAndSubmit(
+			t, b, tx,
+			[]flow.Address{b.ServiceKey().Address, tokenAddr}, []crypto.Signer{b.ServiceKey().Signer(), tokenSigner},
+			false,
+		)
+
+		// Shoulf fail because the price is wrong
```
DavidRS91, line 1043: Minor
```suggestion
		// Should fail because the price is wrong
```

transactions/marketV3/mint_and_purchase.cdc
```diff
@@ -17,8 +18,8 @@ transaction(sellerAddress: Address, recipient: Address, tokenID: UInt64, purchas
         destroy minter
 
         let seller = getAccount(sellerAddress)
-        let topshotSaleCollection = seller.getCapability(/public/topshotSaleCollection)
-            .borrow<&{TopShotMarketV2.SalePublic}>()
+        let topshotSaleCollection = seller.getCapability(TopShotMarketV3.marketPublicPath)
+            .borrow<&{Market.SalePublic}>()
```
DavidRS91, line 22: Should this be v3?

contracts/TopShotMarketV3.cdc
joshuahannan: There isn't logic for it in the smart contract, but we could update the transfer moment transaction to also cancel the sale if it exists

transactions/marketV3/mint_and_purchase.cdc
joshuahannan: No, this is correct. The public interface for the two is exactly the same so we wanted to use it instead of creating a new one

lib/go/templates/template_utils.go
joshuahannan: No, but I was having a lot of difficulty with go modules and dependencies getting CI to pass properly, and this is the only way I was able to figure out how to make it work. 🤷

contracts/TopShotMarketV3.cdc
joshuahannan: Do you think that you'd like to use a transaction like that?

transactions/marketV3/mint_and_purchase.cdc
DavidRS91: Got it. Thanks for the explanation!

contracts/TopShotMarketV3.cdc
DavidRS91: I think that could make sense - I'll defer to @sadief @EricLin2004 and @rrrkren here though.

DavidRS91 (approved):
LGTM!

contracts/TopShotMarketV3.cdc
joshuahannan: We don't borrow a capability to the moment itself, so nothing would happen to a capability because it doesn't exist. If the moment that was for sale is transferred, then when someone tries to purchase it, the purchase would fail and would revert. If we just put the cancel sale in the transfer transaction, then we would avoid this in the top shot marketplace

contracts/TopShotMarketV3.cdc
```diff
@@ -0,0 +1,345 @@
...
+    Description: Contract definitions for users to sell their moments
+
+    Marketplace is where users can create a sale collection that they
+    store in their account storage. In the sale collection, 
+    they can put their NFTs up for sale with a price and publish a 
+    reference so that others can see the sale.
+
+    If another user sees an NFT that they want to buy,
+    they can send fungible tokens that equal or exceed the buy price
+    to buy the NFT.  The NFT is transferred to them when
```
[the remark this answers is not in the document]
joshuahannan: That kind of introduces more complications. They'd also have to include a capability for their vault so the leftovers would be sent back to them. Also, as far as I know, we don't allow the users to change their prices without first canceling the sale, so I don't think we would ever encounter this

contracts/TopShotMarketV3.cdc
```diff
@@ -0,0 +1,345 @@
...
+            } else if let v1Market = self.marketV1Capability {
+                let v1MarketRef = v1Market.borrow()!
+
+                return <-v1MarketRef.purchase(tokenID: tokenID, buyTokens: <-buyTokens)
+            }
+            
+            panic("No token matching this ID for sale!")
+        }
+
+        /// changePercentage changes the cut percentage of the tokens that are for sale
```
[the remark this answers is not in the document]
joshuahannan: Yeah, the cutPercentage is taken at purchase time, so if someone changes the percentage, it will apply to all the tokens in the sale. I thought we had decided that that was what we wanted. Is it not desired any more?

contracts/TopShotMarketV3.cdc
EricLin2004: We should talk about this. Thanks for clarifying on specifics but the fact remains that we will have some listings in the marketplace that are no longer valid if we don't remove them, yes?

contracts/TopShotMarketV3.cdc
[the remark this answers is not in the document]
EricLin2004: Yeah that's what I was pointing out, your comment here says:

>If another user sees an NFT that they want to buy,
>they can send fungible tokens that equal or exceed the buy price

Says equal or exceeds.

contracts/TopShotMarketV3.cdc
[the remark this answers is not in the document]
EricLin2004: I don't remember why we set it this way though, do you?

It feels odd that you can change the cutPercentage of EXISTING listed items.

A malicious operator could wait for everyone to list and say the cutPercentage is 1% but after everything is listed they can set it to 50% or something?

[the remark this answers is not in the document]
rrrkren: maybe add a check in purchase to validate cut percentage?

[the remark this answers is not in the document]
rrrkren: what exactly does that achieve thou?
buyer: they pay the same amount regardless of cut percentage
seller: they would be the party that authorize the cut percentage change?

transactions/user/transfer_moment_v3_sale.cdc
```diff
@@ -0,0 +1,43 @@
...
+        // borrow a reference to the owner's collection
+        let collectionRef = acct.borrow<&TopShot.Collection>(from: /storage/MomentCollection)
+            ?? panic("Could not borrow a reference to the stored Moment collection")
+        
+        // withdraw the NFT
+        self.transferToken <- collectionRef.withdraw(withdrawID: withdrawID)
+
+        if let saleRef = acct.borrow<&TopShotMarketV3.SaleCollection>(from: TopShotMarketV3.marketStoragePath) {
+            saleRef.cancelSale(tokenID: withdrawID)
+        }
```
sadief, line 29: Are we already cancelling the sale before a transaction here?

joshuahannan: Yeah, this is transaction that cancels the sale at the same time as the transfer

contracts/TopShotMarketV3.cdc
sadief: Can I clarify - The `cancelSale` already happens upon a transaction [here](https://github.com/dapperlabs/nba-smart-contracts/pull/118#discussion_r666547298) so confirming that there is no action required on this comment now? @EricLin2004 @joshuahannan

contracts/TopShotMarketV3.cdc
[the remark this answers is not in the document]
sadief: @EricLin2004 @rrrkren @joshuahannan clarification: do we want to remove `changePercentage` entirely?

contracts/TopShotMarketV3.cdc
joshuahannan: Yes, that is correct

contracts/TopShotMarketV3.cdc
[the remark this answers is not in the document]
joshuahannan: Yeah, the seller is the one who sets the cut percentage, so it would only change it for their sales, which would only affect them. The buyer still pays the same price. I think we should keep the `changePercentage` function in case the top shot team decides to change it in the future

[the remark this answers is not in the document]
sadief: Just chatted with @EricLin2004 - he would like the `changePercentage` to be removed as we don't currently use it and likely will not (can't think of any reason that it would make sense for a seller to be able to change the cut percentage) and since the contracts are public we don't want to set a precedent for it that others may use as an example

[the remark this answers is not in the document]
sadief: Removed `changePercentage`

contracts/TopShotMarketV3.cdc
```diff
@@ -0,0 +1,345 @@
...
+        /// Parameters: tokenID: the ID of the NFT to purchase
+        ///             buyTokens: the fungible tokens that are used to buy the NFT
+        ///
+        /// Returns: @TopShot.NFT: the purchased NFT
+        pub fun purchase(tokenID: UInt64, buyTokens: @FungibleToken.Vault): @TopShot.NFT {
+
+            // First check this sale collection for the NFT
+            if self.prices[tokenID] != nil {
+                assert(
+                    buyTokens.balance == self.prices[tokenID]!,
```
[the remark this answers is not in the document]
sadief: Resolved - removed comment

EricLin2004 (approved):

rrrkren (approved):

PR #125, opened 2021-09-14
Improvements to the Top Shot Contract

joshuahannan:
This draft includes my proposed changes to `TopShot.cdc` to fix a few issues

First issue:
## Make array and dictionary fields private
Public array and dictionary fields cannot be directly assigned to, but their elements can be modified, so I have made them private so the admin cannot modify these fields

## Create a more easily Queryable struct for set metadata
Currently, querying for set data is difficult because you have to do individual function calls. This method puts all the important data into a single struct so it can be queried all at once

### DIFF
--- a/contracts/TopShot.cdc
+++ b/contracts/TopShot.cdc
@@ -143,7 +143,7 @@ pub contract TopShot: NonFungibleToken {
         // This is not the long term way NFT metadata will be stored. It's a temporary
         // construct while we figure out a better way to do metadata.
         //
-        pub let metadata: {String: String}
+        access(self) let metadata: {String: String}
 
         init(metadata: {String: String}) {
             pre {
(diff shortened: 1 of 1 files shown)


renschler:
This might not be an appropriate place to ask this... but is there a way to tell on github if this change has been deployed to mainnet? Along the same lines, when the change is live, does the contract live at the same address? Or does flow require that you create a new contract with a new address?

joshuahannan:
@renschler Great question! We don't have a current indication on github if the contracts are deployed, but that is something we could definitely add. Great suggestion. If a project doesn't have an indication like that though, you could always just [check the address on chain](https://flow-view-source.com/mainnet/account/0x0b2a3299cc857e29/contract/TopShot) to see if the code is the same. We have deployed this upgrade though, so the update in the PR is on-chain.

Upgrades can be done on the same address, so we didn't have to re-deploy for this one. There are restrictions for upgrades though: https://docs.onflow.org/cadence/language/contract-updatability/

PR #128, opened 2021-09-28
Add events from nba api with unit tests

Deewai:


lib/go/events/deposit.go
```diff
@@ -37,10 +38,18 @@ func (evt depositEvent) Owner() string {
 	return evt.To()
 }
 
+func (evt depositEvent) isValidEvent() bool {
```
rrrkren, line 41: would a err response be more versatile than bool?

fkenji (approved):
This is great! 👍

lib/go/events/deposit.go
Deewai: there is an error here https://github.com/dapperlabs/nba-smart-contracts/pull/128/files#diff-c28b4f3f5872bedc52dcc2b8e8d09c8a0d1fa9a4b990cf12f1fd1ff06f9043c2R52 and since this is not an exported method, i don't think it is needed

lib/go/events/set_play_retired.go
```diff
@@ -0,0 +1,49 @@
...
+	"fmt"
+	"github.com/onflow/cadence"
+	jsoncdc "github.com/onflow/cadence/encoding/json"
+)
+
+var (
+	EventPlayRetiredFromSet = "TopShot.PlayRetiredFromSet"
+)
+
+type SetPlayRetiredEvent interface {
```
jordancoblin, line 13: Out of scope for this PR, but I'm not sure why these needed to be interfaces. Seems like we could have just used a struct type, e.g.

```golang
type SetPlayRetiredEvent struct {
    rawEvent cadence.Event
}

func (evt SetPlayRetiredEvent) SetID() uint32 {
	return uint32(evt.Fields[0].(cadence.UInt32))
}

etc...
```

joshuahannan (approved):

lib/go/events/set_play_retired.go
jordancoblin: Or actually better:

```
type SetPlayRetiredEvent struct {
    SetID uint32
    PlayID uint32
    NumMoments uint32
}
```

And then done all the mapping in the `Decode` method.

jordancoblin (approved):
Awesome, thanks for this 🙏

sadief (approved):
🙌

lib/go/events/deposit.go
rrrkren: if we return an error here and rename to `validate() error`, then we could potentially surface more information about the error?
```go
fmt.Errorf("error decoding event: event is not a valid deposit event, expected type %s, got %s",expected,got)
```

Deewai: Done

PR #129, opened 2021-10-13
Add revealed event

qhou:
- Add revealed event decoding and unit tests
- Smart contract from PDS for revealed event https://github.com/flow-hydraulics/flow-pds/blob/main/cadence-contracts/IPackNFT.cdc#L27


qhou:
@joshuahannan  can you merge this if it's good to go? I'm not authorized

Issue #133, opened 2021-11-12
Update README to align with latest improvements


joshuahannan:
### Issue To Be Solved
The Top Shot contract was recently upgraded and the docs need to be updated.

### Suggest A Solution
* Update the README to mention the changes where applicable from this PR: https://github.com/dapperlabs/nba-smart-contracts/pull/125
* Give examples for querying set data
* https://blog.nbatopshot.com/posts/nba-top-shot-smart-contract-improvements

PR #135, opened 2021-12-08
Add contract improvements to readme

Deewai:
Closes https://github.com/dapperlabs/nba-smart-contracts/issues/133


joshuahannan (commented):
Thank you for submitting this! I should have been more clear in the issue, but I mostly meant that I wanted to make sure that the information that was already in the README was still correct after the upgrade, and that there were instructions for `QuerySetData` were included. 

I think it is still good to mention that there was an upgrade and link to the blog post, but I don't think there needs to be details about each part of the upgrade except the for set metadata part. 

Does that make sense?

Deewai:
> Thank you for submitting this! I should have been more clear in the issue, but I mostly meant that I wanted to make sure that the information that was already in the README was still correct after the upgrade, and that there were instructions for `QuerySetData` were included.
> 
> I think it is still good to mention that there was an upgrade and link to the blog post, but I don't think there needs to be details about each part of the upgrade except the for set metadata part.
> 
> Does that make sense?

@joshuahannan how about now?

joshuahannan (approved):
Looks good! Thanks for making the changes. :)

Issue #134, opened 2021-12-03
Update `lib/go/test` package to use latest flow and Cadence dependencies


joshuahannan:
### Issue To Be Solved
Need to use the latest versions of all the dependencies in the `lib/go/test` package

### Suggest A Solution
* Update to [Cadence v0.20.1](https://github.com/onflow/cadence/releases/tag/untagged-a1317689821ee2384185)
* Update to [Flow-Go-SDK v0.24.0](https://github.com/onflow/flow-go-sdk/releases/tag/v0.24.0)
* Updated to [Flow Emulator v0.27.0](https://github.com/onflow/flow-emulator/releases/tag/v0.27.0)

PR #136, opened 2021-12-09
Update versions in lib/go/test

HouseOfHufflepuff:
Update cadence, flow-sdk, flow-emulator versions in lib/go/test package per:
https://github.com/dapperlabs/nba-smart-contracts/issues/134


lib/go/test/topshot_test.go
```diff
@@ -398,7 +398,7 @@ func TestMintNFTs(t *testing.T) {
 		assert.Equal(t, cadence.NewBool(true), result)
 
 		result = executeScriptAndCheck(t, b, templates.GenerateGetCollectionIDsScript(env), [][]byte{jsoncdc.MustEncode(cadence.Address(topshotAddr))})
-		idsArray := cadence.NewArray([]cadence.Value{cadence.NewUInt64(1), cadence.NewUInt64(2), cadence.NewUInt64(3), cadence.NewUInt64(4), cadence.NewUInt64(5), cadence.NewUInt64(6)})
+		idsArray := cadence.NewArray([]cadence.Value{cadence.NewUInt64(6), cadence.NewUInt64(4), cadence.NewUInt64(2), cadence.NewUInt64(1), cadence.NewUInt64(5), cadence.NewUInt64(3)})
```
HouseOfHufflepuff, line 401: @joshuahannan result is now returning as [6, 4, 2, 1, 5, 3] each run. Explored assert.Contains but that was not playing well with cadence.value. Any suggestions appreciated, or perhaps this is the only way to do it.

joshuahannan: I think the latest version of Cadence had a change to how arrays are ordered, or something like that... I think this is fine.

joshuahannan (approved):
Thanks for doing this! There actually was another emulator release yesterday: Flow Emulator v0.27.0
Could you update to that one instead? The cadence and flow-go-sdk versions will stay the same.

joshuahannan:
Thanks! Merging now

PR #145, opened 2022-05-27
add autograph fields to play metadata

yeukovichd:
Added autograph fields to play metadata

### DIFF
--- a/lib/go/templates/data/play_metadata.go
+++ b/lib/go/templates/data/play_metadata.go
@@ -29,6 +29,9 @@ type PlayMetadata struct {
 	AwayTeamName         string
 	HomeTeamScore        *int32 `json:",string"`
 	AwayTeamScore        *int32 `json:",string"`
+	AutographType        string `json:",omitempty"`
+	AutographDate        string `json:",omitempty"`
+	AutographSigner      string `json:",omitempty"`
 }
 
 // GenerateEmptyPlay generates a play with all its fields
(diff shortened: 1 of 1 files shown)


joshuahannan:
@yeukovichd Are you part of the Top Shot team? This metadata addition would need to be something that is approved by the core team.

yeukovichd:
> @yeukovichd Are you part of the Top Shot team? This metadata addition would need to be something that is approved by the core team.

@joshuahannan Yes, I am. We are working on new feature that allows to create moments with player signature. I can't pass CI check. Something with invalid contract address. Does your pr https://github.com/dapperlabs/nba-smart-contracts/pull/146 fix it?

joshuahannan:
@yeukovichd You need to run `make test` from the top level of the directory to build all the transactions for the tests. Also, can you update the top shot metadata view in `TopShot.cdc` to include the new fields?

tpetrychyn:
> @yeukovichd You need to run `make test` from the top level of the directory to build all the transactions for the tests. Also, can you update the top shot metadata view in `TopShot.cdc` to include the new fields?

Hey, I'm from Fandom pod Top Shot team. I'm curious why we need to update the contract? Our intent for these fields is just to quash any concerns from users that the signatures are not "on-chain". We don't have any plans to ever read these fields in the Top Shot product. This is only to be used for Magic Johnson's autograph play and is purely a marketing experiment.

If the user reaction is positive, our medium-term plan is to come up with a proper autograph contract that can add an autograph to existing moments. If/when we get a proper autograph contract we will retroactively autograph the Magic Johnson moments and remove/ignore these fields.

Does that change anything?

joshuahannan:
@tpetrychyn The purpose of those views is to return all the metadata about an NFT. Even though that signature field will only be used for the magic johnson moments, we still should be returning that metadata because it is important for those moments and [important to show the full set of metadata for every NFT.](https://github.com/onflow/flow-nft#nft-metadata)

joshuahannan (approved):

PR #148, opened 2022-06-02
Secure cadence updates

fkenji:
ref https://forum.onflow.org/t/breaking-changes-coming-with-secure-cadence-release/3052/7


lib/go/test/topshot_test.go
```diff
@@ -464,7 +464,7 @@ func TestMintNFTs(t *testing.T) {
 		assert.Equal(t, cadence.NewBool(true), result)
 
 		result = executeScriptAndCheck(t, b, templates.GenerateGetCollectionIDsScript(env), [][]byte{jsoncdc.MustEncode(cadence.Address(topshotAddr))})
-		idsArray := cadence.NewArray([]cadence.Value{cadence.NewUInt64(6), cadence.NewUInt64(5), cadence.NewUInt64(1), cadence.NewUInt64(4), cadence.NewUInt64(3), cadence.NewUInt64(2)})
+		idsArray := cadence.NewArray([]cadence.Value{cadence.NewUInt64(2), cadence.NewUInt64(1), cadence.NewUInt64(5), cadence.NewUInt64(4), cadence.NewUInt64(3), cadence.NewUInt64(6)})
```
fkenji, line 467: Changed the ordering of the array here, shouldn't be ok I think - it still contains the same elements

Deewai: yep, the ordering changes a lot. Not sure why

lib/go/events/go.mod
```diff
@@ -3,7 +3,20 @@ module github.com/dapperlabs/nba-smart-contracts/lib/go/events
...
+	github.com/rogpeppe/go-internal v1.8.0 // indirect
+	github.com/sirupsen/logrus v1.6.0 // indirect
+	github.com/stretchr/objx v0.2.0 // indirect
+	github.com/stretchr/testify v1.7.1-0.20210824115523-ab6dc3262822
+	go.uber.org/goleak v1.1.11-0.20210813005559-691160354723 // indirect
+	golang.org/x/crypto v0.0.0-20211215165025-cf75a172585e // indirect
+	golang.org/x/mod v0.5.0 // indirect
+	golang.org/x/tools v0.1.7 // indirect
+	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
+	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
```
joshuahannan, line 21: does this need a `go mod tidy` ?

joshuahannan (approved):
Thanks for making these changes!

sadief (commented):
ooh does this mean that the Market contracts didn't require updates? 🤞

lib/go/test/topshot_test.go
fkenji: It wasn't on the previous version 😓

PR #150, opened 2022-06-23
Feat: Moment Locking

tpetrychyn:
ADR: https://www.notion.so/dapperlabs/ADR-Moment-Locking-9bcf1a46d0564a7ab6ca0c0d56fa30fa#8f092cc9363442c08a7eb4326d9bd1b0

Adds a moment locking smart contract which accepts topshot nfts and stores their ids in a dictionary. After the lock time has passed or an admin has overridden the lock an account can call the unlock function and the id is removed from the list.

Changes:
- New TopShotLocking contract
- TopShot contract exposes private funcs lock/unlock/batchLock/batchUnlock on Collection resource
- TopShot contract `withdraw` now checks the locking contract
- TopShotMarketV3 `listForSale` checks the locking contract


README.md
```diff
@@ -422,6 +422,22 @@ but it will be deployed and utilized in the near future.
 Some improvements were made to the Topshot contract to reflect some cadence best practices and fix a bug.
 In-depth explanation on the changes and why we made them can be found in our [Blog Post](https://blog.nbatopshot.com/posts/nba-top-shot-smart-contract-improvements) 
 
+## TopShot Locking Contract Overview
+
```
joshuahannan, line 426: Add the name of the contract somewhere in here.

Also, mention some specific functions and transactions that can be used to lock/unlock moments and the process that the user should go through to do it.

contracts/TopShot.cdc
```diff
@@ -83,6 +84,12 @@ pub contract TopShot: NonFungibleToken {
     // Emitted when a Moment is destroyed
     pub event MomentDestroyed(id: UInt64)
 
+    // Emitted when a Moment is locked
+    pub event MomentLocked(id: UInt64, duration: UFix64, expiryTimestamp: UFix64)
```
joshuahannan, line 88: I don't think that these events should be defined or used in the TopShot contract. I think they should be only in the locking contract

contracts/TopShot.cdc
```diff
@@ -866,6 +879,51 @@ pub contract TopShot: NonFungibleToken {
...
+
+            // Add the new token to the dictionary
+            let oldToken <- self.ownedNFTs[id] <- TopShotLocking.lockNFT(nft: <- token, expiryTimestamp: lockExpiryTimestamp)
+
+            emit MomentLocked(id: id, duration: duration, expiryTimestamp: lockExpiryTimestamp)
+
+            destroy oldToken
+        }
+
+        pub fun batchLock(ids: [UInt64], duration: UFix64) {
```
joshuahannan, line 898: Add comments here

contracts/TopShotLocking.cdc
```diff
@@ -0,0 +1,114 @@
+import NonFungibleToken from 0xNFTADDRESS
+
+pub contract TopShotLocking {
```
joshuahannan, line 3: Did you decide if you want this contract to be a general utility contract or not? If so, it should be something like TopShotUtilities or something like that

contracts/TopShotLocking.cdc
```diff
@@ -0,0 +1,114 @@
...
+
+    // -----------------------------------------------------------------------
+    // TopShotLocking contract Events
+    // -----------------------------------------------------------------------
+
+    // Emitted when a moment is withdrawn from a Collection
+    pub event Locked(id: UInt64, expiryTimestamp: UFix64)
```
joshuahannan, line 10: should the duration be in this also?

I would also consider adding some other identifying information to these events, such as the playID or setID.

contracts/TopShotLocking.cdc
```diff
@@ -0,0 +1,114 @@
...
+    // Emitted when a moment is deposited into a Collection
+    pub event Unlocked(id: UInt64)
+
+    // Dictionary of locked NFTs
+    // nft resource uuid is the key
+    // locked until timestamp is the value
+    pub var lockedNFTs: {UInt64: UFix64}
```
joshuahannan, line 17:
```suggestion
    access(self) var lockedNFTs: {UInt64: UFix64} 
```

contracts/TopShotLocking.cdc
```diff
@@ -0,0 +1,114 @@
...
+    // Emitted when a moment is deposited into a Collection
+    pub event Unlocked(id: UInt64)
+
+    // Dictionary of locked NFTs
+    // nft resource uuid is the key
+    // locked until timestamp is the value
+    pub var lockedNFTs: {UInt64: UFix64} 
+
+    // Dictionary of NFTs overridden to be unlocked
+    pub var unlockableNFTs: {UInt64: Bool} // nft resource uuid is the key
```
joshuahannan, line 20:
```suggestion
    access(self) var unlockableNFTs: {UInt64: Bool} // nft resource uuid is the key
```

contracts/TopShotLocking.cdc
```diff
@@ -0,0 +1,114 @@
...
+    pub fun lockNFT(nft: @NonFungibleToken.NFT, expiryTimestamp: UFix64): @NonFungibleToken.NFT {
+        let TopShotNFTType: Type = CompositeType("A.0xTOPSHOTADDRESS.TopShot.NFT")!
+        if !nft.isInstance(TopShotNFTType) {
+            panic("NFT is not a TopShot NFT")
+        }
+
+        let id = nft.uuid
+        if self.lockedNFTs.containsKey(id) {
+            panic("NFT is already locked")
+        }
```
joshuahannan, line 45: probably don't need to panic here. You can just skip the locking and event if it is already locked

contracts/TopShotLocking.cdc
```diff
@@ -0,0 +1,114 @@
...
+    // Parameters: nft: NFT resource
+    //
+    // Returns: the NFT resource
+    //
+    // NFT must be eligible for unlocking by an admin
+    pub fun unlockNFT(nft: @NonFungibleToken.NFT): @NonFungibleToken.NFT {
+        let id = nft.uuid
+        if !self.lockedNFTs.containsKey(id) {
+            panic("NFT is not locked")
+        }
```
joshuahannan, line 65: probably don't need to panic here. you can just return the NFT

contracts/TopShotLocking.cdc
```diff
@@ -0,0 +1,114 @@
...
+        self.lockedNFTs.remove(key: id)
+
+        emit Unlocked(id: id)
+
+        return <- nft
+    }
+
+    // Admin is a special authorization resource that 
+    // allows the owner to perform important functions to modify the 
+    // various aspects of the Plays, Sets, and Moments
```
joshuahannan, line 86: update this comment to refer to this contract's Admin functionality

contracts/TopShotLocking.cdc
```diff
@@ -0,0 +1,114 @@
...
+    //
+    pub resource Admin {
+        // createNewAdmin creates a new Admin resource
+        //
+        pub fun createNewAdmin(): @Admin {
+            return <-create Admin()
+        }
+
+        pub fun markNFTUnlockable(nftRef: &NonFungibleToken.NFT) {
```
joshuahannan, line 95: Add a description for this function

lib/go/test/market_test.go
```diff
@@ -62,9 +62,25 @@ func TestMarketDeployment(t *testing.T) {
 	_, err = b.CommitBlock()
 	assert.NoError(t, err)
 
+	// Should be able to deploy the TopShotLocking contract
+	// as a new account with no keys.
+	topShotLockingCode, _ := DownloadFile(TopShotLockingContractsBaseURL + TopShotLockingContractsInterfaceFile)
```
joshuahannan, line 67: Instead of `DownloadFile`, you should be able to add a new method to the `contracts` package (`GenerateTopShotLockingContract`) and use that to get the contract code. We usually try to avoid downloading directly.

transactions/admin/mark_moment_unlockable.cdc
```diff
@@ -0,0 +1,25 @@
+import NonFungibleToken from 0xNFTADDRESS
+import TopShot from 0xTOPSHOTADDRESS
+import TopShotLocking from 0xTOPSHOTLOCKINGADDRESS
+
```
joshuahannan, line 4: Can you add high-level descriptions for all these transactions and some more comments in the code? Thanks!

contracts/TopShot.cdc
tpetrychyn: The reason I defined it here is because in TopShotLocking contract it only knows the UUID of the NFT, but when listening in `nba-api` we need the `TopShot.NFT.id`

joshuahannan: You can also get the id in there. it is a field on the NFT

joshuahannan: It is just `nft.id`

contracts/TopShotLocking.cdc
tpetrychyn: Since the TopShotLocking contract doesn't import the TopShot contract it makes more sense to me that it is "dumb" to all of the inner-TopShot NFT details.

contracts/TopShotLocking.cdc
tpetrychyn: I prefer to make it specific Locking for simplicity

contracts/TopShot.cdc
tpetrychyn: Thanks, I've moved the events to TopShotLocking!

joshuahannan (commented):
The contract looks good to me so far. Just need the tests now

joshuahannan (commented):
Looks good! Can you add a tests to make sure that non-top shot NFT cannot be locked?

lib/go/test/locking_test.go
```diff
@@ -0,0 +1,421 @@
...
+	"github.com/onflow/flow-go-sdk/test"
+	"github.com/stretchr/testify/assert"
+	"strings"
+	"testing"
+	"time"
+)
+
+const CadenceUFix64Factor = 100000000
+
+// Tests all the main functionality of the V1 Market
```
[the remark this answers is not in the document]
tpetrychyn: I have added a test to ensure only official TopShot.NFTs are allowed to be locked

contracts/TopShotLocking.cdc
```diff
@@ -0,0 +1,135 @@
...
+        return self.lockedNFTs[nftRef.uuid]!
+    }
+
+    // lockNFT Takes an NFT resource and adds its unique identifier to the lockedNFTs dictionary
+    //
+    // Parameters: nft: NFT resource
+    //             expiryTimestamp: The unix timestamp in seconds after which the nft may be unlocked
+    //
+    // Returns: the NFT resource
+    pub fun lockNFT(nft: @NonFungibleToken.NFT, duration: UFix64): @NonFungibleToken.NFT {
```
Deewai, line 50: is there a reason why the nft needs to be passed to this function?.  you should be able to get the information you need by passing a reference of the nft. That way you don't need to emit deposit events that changes nothing

contracts/TopShotLocking.cdc
```diff
@@ -0,0 +1,135 @@
...
+        let TopShotNFTType: Type = CompositeType("A.TOPSHOTADDRESS.TopShot.NFT")!
```
Deewai, line 51: i don't really like the idea of having template variables in the contract code

Deewai: have you considered having this contract setup in each user's account and the nftType can be passed to the function. This way it can be reusable by other SC. Then `lockedNFTs` is of type `{Type:{UInt64: UFix64}}`

lib/go/templates/template_utils.go
```diff
@@ -21,6 +21,8 @@ const (
 	placeholderDUCAddress             = "0xDUCADDRESS"
 	placeholderForwardingAddress      = "0xFORWARDINGADDRESS"
 	placeholderMetadataViewsAddress   = "0xMETADATAVIEWSADDRESS"
+	placeholderTopShotLockingAddress  = "0xTOPSHOTLOCKINGADDRESS"
+	placeholderFakeTopShotAddress     = "0xFAKETOPSHOTADDRESS"
```
Deewai, line 25: what does fakeTopshotAddress do?

transactions/admin/fake_admin/add_play_to_set.cdc
```diff
@@ -0,0 +1,36 @@
+import TopShot from 0xFAKETOPSHOTADDRESS
```
Deewai, line 1: what makes the address fake?

Deewai (commented):
Left a couple of comments, but great work 👏

contracts/TopShotLocking.cdc
tpetrychyn: The nft resource must be sent to this contract to prove ownership, otherwise I could simply call `borrowNFT` on anyone's moments and lock them

lib/go/templates/template_utils.go
tpetrychyn: fake topshot address is used by the last test in `locking_test`: "Should not be able to lock a non-TopShot.NFT". It is used in testing to deploy a second copy of the TopShot address as a "fake or inauthentic" TopShot contract. I then mint moments from it and try to lock them. It exercises the check that the locking contract will only hold TopShot.NFTs.

contracts/TopShotLocking.cdc
tpetrychyn: This relates to a discussion @joshuahannan and I had where we wanted to prevent someone from "spamming" the contract with non-topshot nfts and potentially slowing down the contract as the dictionary fills up.

Also, given that there is an admin backdoor which can unlock nfts prematurely, it makes the most sense to scope this purely for TopShot NFTs.

transactions/admin/fake_admin/add_play_to_set.cdc
tpetrychyn: See previous comment.

contracts/TopShotLocking.cdc
Deewai: ha, that makes sense 👍

contracts/TopShotLocking.cdc
Deewai: i see. Good point

lib/go/templates/template_utils.go
Deewai: ha, nice test

transactions/admin/fake_admin/create_set.cdc
```diff
@@ -0,0 +1,35 @@
+import TopShot from 0xFAKETOPSHOTADDRESS
```
Deewai, line 1: it is my understanding that the only thing different from the tx and the original create set tx is the Topshot address. If this is the case we shouldn't need to write a new tx file. Yo just need to replace the address with the fake address

Deewai: same applies for other txs

Deewai (approved):
Left a couple of non-blocking comments

transactions/admin/fake_admin/create_set.cdc
tpetrychyn: Good point, I have removed the fake topshot address and instead I just use a modified env for the tests

contracts/TopShotMarketV3.cdc
```diff
@@ -138,6 +139,9 @@ pub contract TopShotMarketV3 {
             pre {
                 self.ownerCollection.borrow()!.borrowMoment(id: tokenID) != nil:
                     "Moment does not exist in the owner's collection"
+
```
sadief, line 142: I think you will need to add this to the [MarketTopShotOldVersion.cdc](https://github.com/dapperlabs/nba-smart-contracts/blob/master/contracts/MarketTopShotOldVersion.cdc) as well? since NBA has to support both contracts (If a moment was listed with the old version then the V3 checks it and processes either the delisting or purchasing with the old contract)

sadief (commented):
Looks good! One small comment about desired behaviour and I think we need an update to the old market contract as well

contracts/TopShotMarketV3.cdc
sadief: Ohh yes, cool, no update needed then 👍

contracts/TopShot.cdc
```diff
@@ -866,6 +873,52 @@ pub contract TopShot: NonFungibleToken {
...
+
+            destroy oldToken
+        }
+
+        // batchUnlock takes an array of token ids
+        // it iterates through the ids and unlocks each if they are eligible
+        pub fun batchUnlock(ids: [UInt64]) {
+            // Iterate through the ids and unlocks them
+            for id in ids {
+                self.unlock(id: id)
```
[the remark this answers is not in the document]
sadief: Rollback by default would happen in the TX - resolving!

sadief (approved):

HouseOfHufflepuff (approved):
Nice work.

joshuahannan (approved):
Is this deployed yet? Can the PR be merged?

PR #151, opened 2022-07-13
Fix: TopShot Locking use nft.id for lock key

tpetrychyn:
There is a bug in cadence where UUID is not unique: https://axiomzen.slack.com/archives/CG0B7CJAJ/p1657750983733739

This changes the contract to use id

Also adds admin endpoint to unlock all moments since we need to cleanup testnet and mainnet.


jrkhan (approved):

joshuahannan:
Are we sure this is actually a bug yet? From the thread in slack, it looks to me like it was just an issue with resources that were created in summer 2020. We might not want to make this change until that is confirmed

tpetrychyn:
> Are we sure this is actually a bug yet? From the thread in slack, it looks to me like it was just an issue with resources that were created in summer 2020. We might not want to make this change until that is confirmed

Yep, we caught 2 moments that collided during QA. A moment that was never locked was unable to be listed for sale with error, "moment locked". 

Running the following query in production returns 90,000 moments that were minted before this bug was fixed:
`select count(*) from moments where created_at < '2020-07-01 00:32:51.754798'`

Realistically the contract should have used `nft.id` as soon as we decided to scope it to TopShot.NFTs.

The contract is already deployed to testnet and mainnet because the launch is tomorrow morning..

joshuahannan (approved):
ok sounds good

PR #152, opened 2022-07-22
Add core NFT views placeholder

jrkhan:
For discussion, an initial implementation of core NFT views. 
See: https://www.flow-nft-catalog.com/

This should also allow the image url associated with the moment to be resolved on chain.

Notes:
This makes the assumption that the following play metadata will be added (separate PR to the topshot repo forthcoming):
DefaultSetSlug -> The "SetSlug" of the play, currently every play will only have one, but if we end up minting a play into multiple sets we may need to allow per moment overrides for this. 
PlayGuid -> Currently only stored off chain, part of the image URL
Two alternatives to ^ as building image urls on chain dynamically seems out of places.
* Deploy a reverse proxy/gateway/simply another bucket to provide a more ergonomic path for each moment/play (perhaps just requiring the on chain momentId/playId or content hash)
* Store the precomputed full image url (draw back is if the url changes, we'll need to update each play)

Second, there is a metadata view for an external URL for the moment. While we do not currently have a route
for moment by flow id, it seems preferable to add a route for this instead of adding a moment guid on chain for all 50million+ moments. If this External URL is not required, we may also just omit this metadata view if creating a route by flow id will be problematic in the short term.

TODO:
Once we've decided on appropriate patterns for url creation will add unit tests accordingly

### DIFF
--- a/contracts/TopShot.cdc
+++ b/contracts/TopShot.cdc
@@ -668,8 +668,8 @@ pub contract TopShot: NonFungibleToken {
                     )
                 case Type<MetadataViews.Editions>():
                     let name = self.getEditionName()
-                    let max = TopShot.getNumMomentsInEdition(setID: self.data.setID, playID: self.data.playID) ?? 0;
-                    let editionInfo = MetadataViews.Edition(name: name, number: UInt64(self.data.serialNumber), max: UInt64(max))
+                    let max = TopShot.getNumMomentsInEdition(setID: self.data.setID, playID: self.data.playID) ?? 0
+                    let editionInfo = MetadataViews.Edition(name: name, number: UInt64(self.data.serialNumber), max: max > 0 ? UInt64(max) : nil)
                     let editionList: [MetadataViews.Edition] = [editionInfo]
                     return MetadataViews.Editions(
                         editionList
--- a/contracts/TopShot.cdc
+++ b/contracts/TopShot.cdc
@@ -692,25 +692,35 @@ pub contract TopShot: NonFungibleToken {
                         providerPath: /private/MomentCollection,
                         publicCollection: Type<&TopShot.Collection{TopShot.MomentCollectionPublic}>(),
                         publicLinkedType: Type<&TopShot.Collection{TopShot.MomentCollectionPublic,NonFungibleToken.Receiver,NonFungibleToken.CollectionPublic,MetadataViews.ResolverCollection}>(),
-                        providerLinkedType: Type<&TopShot.Collection{NonFungibleToken.Provider}>(),
+                        providerLinkedType: Type<&TopShot.Collection{NonFungibleToken.Provider,TopShot.MomentCollectionPublic,NonFungibleToken.Receiver,NonFungibleToken.CollectionPublic,MetadataViews.ResolverCollection}>(),
                         createEmptyCollectionFunction: (fun (): @NonFungibleToken.Collection {
                             return <-TopShot.createEmptyCollection()
                         })
                     )
                 case Type<MetadataViews.NFTCollectionDisplay>():
-                    let media = MetadataViews.Media(
+                    let bannerImage = MetadataViews.Media(
                         file: MetadataViews.HTTPFile(
                             url: "https://nbatopshot.com/static/img/top-shot-logo-horizontal-white.svg"
                         ),
                         mediaType: "image/svg+xml"
                     )
+                    let squareImage = MetadataViews.Media(
+                        file: MetadataViews.HTTPFile(
+                            url: "https://nbatopshot.com/static/img/og/og.png"
+                        ),
+                        mediaType: "image/png"
+                    )
                     return MetadataViews.NFTCollectionDisplay(
-                        name: "NBA TopShot",
+                        name: "NBA-Top-Shot",
                         description: "NBA Top Shot is your chance to own, sell, and trade official digital collectibles of the NBA and WNBA's greatest plays and players",
                         externalURL: MetadataViews.ExternalURL("https://nbatopshot.com"),
-                        squareImage: media,
-                        bannerImage: media,
-                        socials: {}
+                        squareImage: squareImage,
+                        bannerImage: bannerImage,
+                        socials: {
+                            "twitter": MetadataViews.ExternalURL("https://twitter.com/nbatopshot"),
+                            "discord": MetadataViews.ExternalURL("https://discord.com/invite/nbatopshot"),
+                            "instagram": MetadataViews.ExternalURL("https://www.instagram.com/nbatopshot")
+                        }
                     )
                 case Type<MetadataViews.Traits>():
                     let traitDictionary: {String: AnyStruct} = {
--- a/contracts/TopShot.cdc
+++ b/contracts/TopShot.cdc
@@ -617,7 +617,8 @@ pub contract TopShot: NonFungibleToken {
                 Type<MetadataViews.NFTCollectionData>(),
                 Type<MetadataViews.NFTCollectionDisplay>(),
                 Type<MetadataViews.Serial>(),
-                Type<MetadataViews.Traits>()
+                Type<MetadataViews.Traits>(),
+                Type<MetadataViews.Medias>()
             ]
         }
 
@@ -629,7 +630,7 @@ pub contract TopShot: NonFungibleToken {
                     return MetadataViews.Display(
                         name: self.name(),
                         description: self.description(),
-                        thumbnail: MetadataViews.HTTPFile(url: self.imageURL())
+                        thumbnail: MetadataViews.HTTPFile(url: self.thumbnail())
                     )
                 // Custom metadata view unique to TopShot Moments
                 case Type<TopShotMomentMetadataView>():
@@ -680,7 +681,7 @@ pub contract TopShot: NonFungibleToken {
                 case Type<MetadataViews.Royalties>():
                     return MetadataViews.Royalties(
                         royalties: [
-                            // No royalty for Moments yet defined for 3rd party marketplaces
+                            // Reserved for 3rd party marketplace royalty
                         ]
                     )
                 case Type<MetadataViews.ExternalURL>():
@@ -732,6 +733,17 @@ pub contract TopShot: NonFungibleToken {
                         "Serial Number": self.data.serialNumber
                     }
                     return MetadataViews.dictToTraits(dict: traitDictionary, excludedNames: [])
+                case Type<MetadataViews.Medias>():
+                    return MetadataViews.Medias(
+                        items: [
+                            MetadataViews.Media(
+                                file: MetadataViews.HTTPFile(
+                                    url: self.mediumimage()
+                                ),
+                                mediaType: "image/jpeg"
+                            )
+                        ]
+                    )
             }
 
             return nil
@@ -744,7 +756,7 @@ pub contract TopShot: NonFungibleToken {
         pub fun getMomentURL(): String {
             return "https://nbatopshot.com/moment/".concat(self.id.toString())
         }
-        // getEditionName Moment's edition name a combination of the Moment's setName and playID
+        // getEditionName Moment's edition name is a combination of the Moment's setName and playID
         // `setName: #playID`
         pub fun getEditionName() : String {
             let setName: String = TopShot.getSetName(setID: self.data.setID) ?? ""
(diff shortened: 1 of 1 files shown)


contracts/TopShot.cdc
```diff
@@ -654,10 +665,95 @@ pub contract TopShot: NonFungibleToken {
...
+                case Type<MetadataViews.ExternalURL>():
+                    return MetadataViews.ExternalURL(self.getMomentURL())
+                case Type<MetadataViews.NFTCollectionData>():
+                    return MetadataViews.NFTCollectionData(
+                        storagePath: /storage/MomentCollection,
+                        publicPath: /public/MomentCollection,
+                        providerPath: /private/MomentCollection,
+                        publicCollection: Type<&TopShot.Collection{TopShot.MomentCollectionPublic}>(),
+                        publicLinkedType: Type<&TopShot.Collection{TopShot.MomentCollectionPublic,NonFungibleToken.Receiver,NonFungibleToken.CollectionPublic,MetadataViews.ResolverCollection}>(),
+                        providerLinkedType: Type<&TopShot.Collection{NonFungibleToken.Provider}>(),
```
aishairzay, line 695: The provider linked type is what type should be linked for private links, and it works best if its a supertype over the publicLinkedType so DAPPs can find all info it needs from the private link if available rather than also having to resolve the public link.

Can we change this to be the following lengthy type:
`Type<&TopShot.Collection{NonFungibleToken.Provider,TopShot.MomentCollectionPublic,NonFungibleToken.Receiver,NonFungibleToken.CollectionPublic,MetadataViews.ResolverCollection}>()`

contracts/TopShot.cdc
```diff
@@ -654,10 +665,95 @@ pub contract TopShot: NonFungibleToken {
...
+                        ),
+                        mediaType: "image/svg+xml"
+                    )
+                    return MetadataViews.NFTCollectionDisplay(
+                        name: "NBA TopShot",
+                        description: "NBA Top Shot is your chance to own, sell, and trade official digital collectibles of the NBA and WNBA's greatest plays and players",
+                        externalURL: MetadataViews.ExternalURL("https://nbatopshot.com"),
+                        squareImage: media,
+                        bannerImage: media,
+                        socials: {}
```
aishairzay, line 713: Would be nice to have these filled in to topshot's discord and twitter.

It'd look like the following:
```
{
  "twitter":  MetadataViews.ExternalURL("https://twitter.com/nbatopshot"),
  "discord": ...
}
```

aishairzay (commented):
Overall, looks great! Added a few small comments.

If we want to add the topshot moment's video on-chain as well, I'm thinking we could use the `Media` type to portray that in an additional view. https://github.com/onflow/flow-nft/blob/master/contracts/MetadataViews.cdc#L368

contracts/TopShot.cdc
```diff
@@ -654,10 +665,95 @@ pub contract TopShot: NonFungibleToken {
...
+                    )
+                case Type<MetadataViews.Traits>():
+                    let traitDictionary: {String: AnyStruct} = {
+                        "Date Of Moment": TopShot.getPlayMetaDataByField(playID: self.data.playID, field: "DateOfMoment"),
```
austinkline, line 717: Any chance we can find a way to get Tier in here as well? Happy to help contribute to that if it's possible. That seems like the last piece to get all TS data on-chain

jrkhan: Sure thing! Added you to the forked repo if you'd like to add to the PR. 
Currently off chain, moments have a 'tier'..  in all current use cases, all moments for a 'play' will share the tier so it should be safe to add to play metadata for now. 
Previously moments inherited their tier from the _set_ but now have the option to override when they are minted. 
Backfilling tier for all existing plays should be possible (would need a job for this). Also, we'd likely want all newly added moments to add their tier (perhaps a new dictionary of moment id to tier would be appropriate?)

austinkline: > Sure thing! Added you to the forked repo if you'd like to add to the PR.

🙏 If we can get some kind of approval for how Tier could be added, I'm happy to help out there for sure

> This makes the assumption that the following play metadata will be added ....

Since you are already adding data to plays, can Tier just be added as well in one go? It looks like for newly created plays that would be fairly simple, the metadata supplied for the play should just include `Tier`:
https://github.com/dapperlabs/nba-smart-contracts/blob/b6a8ab1611010900c87d3feacadddcc9508d2bbd/contracts/TopShot.cdc#L774

The risky part probably is about adding data to existing plays just in terms of allowing that kind of operation (other pieces of data also need to be added either way), but the method could be quite simple:

```
pub fun addTierToPlay(tier: String, playID: UInt64) {
  TopShot.playDatas[newID]!.metadata["Tier"] = tier
}
```

Then it's just a matter of backfilling which is up to Dapper/TS folks.

contracts/TopShot.cdc
```diff
@@ -654,10 +665,95 @@ pub contract TopShot: NonFungibleToken {
...
+                        editionList
+                    )
+                case Type<MetadataViews.Serial>():
+                    return MetadataViews.Serial(
+                        UInt64(self.data.serialNumber)
+                    )
+                case Type<MetadataViews.Royalties>():
+                    return MetadataViews.Royalties(
+                        royalties: [
+                            // No royalty for Moments yet defined for 3rd party marketplaces
```
austinkline, line 683: Aren't marketplaces having to give a 5% royalty just like TopShot is? With image data being added here, no royalty data could lead some marketplaces to being in trouble. Any idea if/when this might change?

contracts/TopShot.cdc
```diff
@@ -654,10 +665,95 @@ pub contract TopShot: NonFungibleToken {
...
+        pub fun getEditionName() : String {
+            let setName: String = TopShot.getSetName(setID: self.data.setID) ?? ""
+            let editionName = setName.concat(": #").concat(self.data.playID.toString())
+            return editionName
+        }
+
+        // imageURL will build the url for the image associated with the moment based on its metadata
+        //
+        // Returns: The computed external url of a medium sized image associated with the moment
+        pub fun imageURL(): String {
```
austinkline, line 748: Is there a license assumption here that people should be aware of? My guess is that just because an image is here doesn't mean that everyone is allowed to show it. Might be worth adding that into the comments if applicable

contracts/TopShot.cdc
```diff
@@ -654,10 +665,95 @@ pub contract TopShot: NonFungibleToken {
...
+                            url: "https://nbatopshot.com/static/img/top-shot-logo-horizontal-white.svg"
+                        ),
+                        mediaType: "image/svg+xml"
+                    )
+                    return MetadataViews.NFTCollectionDisplay(
+                        name: "NBA TopShot",
+                        description: "NBA Top Shot is your chance to own, sell, and trade official digital collectibles of the NBA and WNBA's greatest plays and players",
+                        externalURL: MetadataViews.ExternalURL("https://nbatopshot.com"),
+                        squareImage: media,
+                        bannerImage: media,
```
bjartek, line 712: I would humbly suggest you use a different image for square/banner. Banner is typically wide and low while square is well square.

contracts/TopShot.cdc
```diff
@@ -654,10 +665,95 @@ pub contract TopShot: NonFungibleToken {
...
+                    )
+                case Type<MetadataViews.NFTCollectionDisplay>():
+                    let media = MetadataViews.Media(
+                        file: MetadataViews.HTTPFile(
+                            url: "https://nbatopshot.com/static/img/top-shot-logo-horizontal-white.svg"
+                        ),
+                        mediaType: "image/svg+xml"
+                    )
+                    return MetadataViews.NFTCollectionDisplay(
+                        name: "NBA TopShot",
```
bjartek, line 708: My itentntion for the Name when we talked about that standard was that it should e able to be used as a slug in an url. This will def work but replacing the space with another seperator or just removing it works better"

contracts/TopShot.cdc
```diff
@@ -654,10 +665,95 @@ pub contract TopShot: NonFungibleToken {
...
+                        description: "NBA Top Shot is your chance to own, sell, and trade official digital collectibles of the NBA and WNBA's greatest plays and players",
+                        externalURL: MetadataViews.ExternalURL("https://nbatopshot.com"),
+                        squareImage: media,
+                        bannerImage: media,
+                        socials: {}
+                    )
+                case Type<MetadataViews.Traits>():
```
bjartek, line 715: is there a particular reason why the rest of the traits from TopShotMomentMetadataView are not added here? If you want thirdparty tools to be able to read this data it is a lot easier for them if you add all of them here.

bjartek: Views that are specific to a producer solution like TopShotMomentMetadataView are really meant for internal consuption. Since using them for external parties will lead to tight couplings and hardcoding rather then exploring and adapting.

contracts/TopShot.cdc
jrkhan: Will use `NBA-Top-Shot`

contracts/TopShot.cdc
```diff
@@ -654,10 +665,95 @@ pub contract TopShot: NonFungibleToken {
...
+                case Type<MetadataViews.NFTCollectionDisplay>():
+                    let media = MetadataViews.Media(
+                        file: MetadataViews.HTTPFile(
+                            url: "https://nbatopshot.com/static/img/top-shot-logo-horizontal-white.svg"
+                        ),
+                        mediaType: "image/svg+xml"
+                    )
+                    return MetadataViews.NFTCollectionDisplay(
+                        name: "NBA TopShot",
+                        description: "NBA Top Shot is your chance to own, sell, and trade official digital collectibles of the NBA and WNBA's greatest plays and players",
```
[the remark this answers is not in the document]
jrkhan: Good catch! Consensus seems to be NBA Top Shot (though we have usage of both on our homepage!)

contracts/TopShot.cdc
jrkhan: The only concern I would have with adding traits for all current fields of TopShotMomentMetadataView would be some of them are optional/will soon be optional (In the near future we may have plays that have no specific player in the metadata, instead the play would be attributed t a whole team). Would there be any concerns with having traits present for some moments and not others?

contracts/TopShot.cdc
jrkhan: This is a good call out - I had been asking around if there were any partnership related concerns with adding image urls on chain and not found immediate concerns. Going to double check with product/partnerships to see if any caveats would be appropriate here.

contracts/TopShot.cdc
jrkhan: Yes - we do collect 5% royalties for offers and the official marketplace. Definitely planning to add that here.

contracts/TopShot.cdc
austinkline: Traits don't have to be uniform! Think of whether a rookie moment should be a trait or not as an example. Putting a trait for the absence of something on that many moments feels like a waste. All that traits are telling you is what this NFT has, not what the complete of traits in an entire collection can be

aishairzay (approved):

contracts/TopShot.cdc
austinkline: Just a note that not having this puts marketplaces in a sticky spot. Now we have to be aware of when something is a TopShot moment and treat its payout differently than others in terms of it's royalty configuration

contracts/TopShot.cdc
jrkhan: Going to defer on adding a rarity trait for the immediate PR (Agree this still seems valuable! But we're going to try to avoid backfilling play data on chain for building image URLs, in part because  we'd like to avoid building in a path that allows us to edit metadata of existing plays)
In a followup PR would like to explore consider adding a new map that would contain the rarity of each moment that won't require us to modify the existing play metadata and have the appropriate granularity of being per moment instead of per play.

contracts/TopShot.cdc
```diff
@@ -654,10 +665,95 @@ pub contract TopShot: NonFungibleToken {
                         setID: self.data.setID,
                         numMomentsInEdition: TopShot.getNumMomentsInEdition(setID: self.data.setID, playID: self.data.playID)
                     )
+                case Type<MetadataViews.Editions>():
+                    let name = self.getEditionName()
+                    let max = TopShot.getNumMomentsInEdition(setID: self.data.setID, playID: self.data.playID) ?? 0;
```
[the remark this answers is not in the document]
jrkhan: Nil probably makes sense - have adjusted.

contracts/TopShot.cdc
jrkhan: Definitely going to follow up on this soon! Just merging this PR for now to a dev/feature branch to enable others to more easily contribute.

contracts/TopShot.cdc
bjartek: So topshot expects to have a royalty cut but it is not in here? Or is this empty because they have no cut on 3rd party markets?

PR #155, opened 2022-08-03
Add Royalties metadata view implementation.

aishairzay:
Something to call out:

I named the env variable `TOPSHOTROYALTYADDRESS`. I did not find somewhere to set env variables per environment, but when being deployed this env variable should be set to the following per network:
testnet b0aa0d8fa90ef9be
mainnet faf0cc52c6e3acaf 

Also worth noting that `MetadataViews.getRoyaltyReceiverPublicPath()` leads to a path that most probably does not have anything set up on the royalty receiving accounts. This is the path where the FTSwitchBoard (https://github.com/onflow/flow-ft/pull/71) should be setup to on some later date. This view is helpful regardless because it allows the address to be pulled out of the capability to help route the funds.


contracts/TopShot.cdc
```diff
@@ -679,9 +680,15 @@ pub contract TopShot: NonFungibleToken {
                         UInt64(self.data.serialNumber)
                     )
                 case Type<MetadataViews.Royalties>():
+                    let royaltyReceiver: Capability<&{FungibleToken.Receiver}> =
+                        getAccount(0xTOPSHOTROYALTYADDRESS).getCapability<&AnyResource{FungibleToken.Receiver}>(MetadataViews.getRoyaltyReceiverPublicPath())
```
jrkhan, line 684: Presumably we'll need to run a tx to add a receiver at this address?

jrkhan (approved):

contracts/TopShot.cdc
aishairzay: I think we should leave it as-is, because it still does allow reference to the address, and will lead to the proper path eventually. There currently aren't any marketplaces using this view yet, and either marketplaces should use the address to find the proper vault or hopefully we'll get the switchboard ready before then (Which is more likely the case.)

One thing we should do - on these topshot royalty receiving accounts, we will need to setup a vault for FUSD, USDC, and Switchboard when available.

joshuahannan (approved):
seems fine to me! Agreed that we should set up the vaults for other fungible tokens on the top shot account as well as the switchboard contract

Issue #149, opened 2022-06-03
Add additional NFT Metadata Views


joshuahannan:
### Issue To Be Solved
Many more metadata views have been added to the standard repo since the original standard was approved and deployed. The Top Shot contract should include the important ones so that its own 3rd party developers can use it and so that it can be a good example for other projects in the Flow ecosystem.

### Suggest A Solution
Check out the latest views that have been added to the metadata standard
* [README](https://github.com/onflow/flow-nft#nft-metadata)
* [Contract](https://github.com/onflow/flow-nft/blob/master/contracts/MetadataViews.cdc#L136-L417)

Add all views to the Top Shot contract that would be useful

### Context

This probably will require a discussion with product managers to figure out how to show certain data on-chain and which data is important.

PR #161, opened 2022-09-08
Feature/core nft views

jrkhan:
This PR will add the core metadata views required for inclusion in the [Flow NFT Catalog](https://www.flow-nft-catalog.com/) 

Resolves:
https://github.com/dapperlabs/nba-smart-contracts/issues/149


contracts/TopShot.cdc
```diff
@@ -606,21 +607,33 @@ pub contract TopShot: NonFungibleToken {
...
+       
+
         pub fun resolveView(_ view: Type): AnyStruct? {
             switch view {
                 case Type<MetadataViews.Display>():
                     return MetadataViews.Display(
                         name: self.name(),
                         description: self.description(),
-                        thumbnail: MetadataViews.HTTPFile(url:"https://ipfs.dapperlabs.com/ipfs/Qmbdj1agtbzpPWZ81wCGaDiMKRFaRN3TU6cfztVCu6nh4o")
+                        thumbnail: MetadataViews.HTTPFile(url: self.thumbnail())
```
Deewai, line 634: nice

contracts/TopShot.cdc
```diff
@@ -654,10 +667,184 @@ pub contract TopShot: NonFungibleToken {
...
+                        publicPath: /public/MomentCollection,
+                        providerPath: /private/MomentCollection,
+                        publicCollection: Type<&TopShot.Collection{TopShot.MomentCollectionPublic}>(),
+                        publicLinkedType: Type<&TopShot.Collection{TopShot.MomentCollectionPublic,NonFungibleToken.Receiver,NonFungibleToken.CollectionPublic,MetadataViews.ResolverCollection}>(),
+                        providerLinkedType: Type<&TopShot.Collection{NonFungibleToken.Provider,TopShot.MomentCollectionPublic,NonFungibleToken.Receiver,NonFungibleToken.CollectionPublic,MetadataViews.ResolverCollection}>(),
+                        createEmptyCollectionFunction: (fun (): @NonFungibleToken.Collection {
+                            return <-TopShot.createEmptyCollection()
+                        })
+                    )
+                case Type<MetadataViews.NFTCollectionDisplay>():
```
Deewai, line 708: Nice!

contracts/TopShot.cdc
```diff
@@ -654,10 +667,184 @@ pub contract TopShot: NonFungibleToken {
...
 
             return nil
         }   
+
+        // Functions used for computing MetadataViews 
+
+        // mapPlayData helps build our trait map from play metadata
+        // Returns: The trait map with all non-nil fields from play data added
+        pub fun mapPlayData(dict: {String: AnyStruct}) : {String: AnyStruct} {      
+            let playMapping: {String: String} = {
```
Deewai, line 770: do we have this just for the display name? It doesn't seem extensible because if we decide to add extra metadata to new plays we mint, we will have to edit this contract. I'm kinda okay with us showing CamelCase for the display name

contracts/TopShot.cdc
```diff
@@ -654,10 +667,184 @@ pub contract TopShot: NonFungibleToken {
...
+        }
+        // getEditionName Moment's edition name is a combination of the Moment's setName and playID
+        // `setName: #playID`
+        pub fun getEditionName() : String {
+            let setName: String = TopShot.getSetName(setID: self.data.setID) ?? ""
+            let editionName = setName.concat(": #").concat(self.data.playID.toString())
+            return editionName
+        }
+
+        pub fun assetPath(): String {
```
Deewai, line 818: Nice!

contracts/TopShot.cdc
```diff
@@ -654,10 +667,184 @@ pub contract TopShot: NonFungibleToken {
...
+
+        // a url to display a video associated with the moment
+        pub fun video(): String {
+            let url = self.assetPath().concat("/video")
+            return self.appendOptionalParams(url: url, firstDelim: "?")
+        }
+
+        // appends and optional network param needed to resolve the media
+        pub fun appendOptionalParams(url: String, firstDelim: String): String {
+            let env = 0xNETWORK
```
Deewai, line 842: i think it makes sense to define these template parameters as top level variables so it doesn't get lost in the contract. At a glance everyone can know the template parameters needed to deploy the contract

Deewai (approved):
LGTM! A couple of non-blocking comments

contracts/TopShot.cdc
jrkhan: Yes, just for display name -> I was going to say we had this also had this map to exclude some fields from getting included as traits, but that can probably better be accomplished by using the 'excludedFields' parameter.
Have updated -> now we'l use the names already present in the metadata map. Double checking, it looks like camel case names are appear in examples of Trait view usage, so I suspect UpperCamelCase should also be okay.

Issue #183, opened 2022-11-30
Ensure sufficient storage before depositing a moment


jrkhan:
### Issue To Be Solved
When a moment is about to be deposited into a custodial wallet, it would be helpful if the transaction not to fail due to insufficient account storage.

### (Optional): Suggest A Solution
One option is to simply check for this in the deposit (see PR) 
Other options, would be to systematically add it to each of our txs -> however there is no guarantee a third party would use our tx.

PR #184, opened 2022-11-30
Try to refill custodial accounts

jrkhan:
One way to approach issue:
https://github.com/dapperlabs/nba-smart-contracts/issues/183

We will check the storage before depositing, this check should fail silently for non dapper accounts, and have the benefit of allowing third parties depositing moments into a custodial account to not need to modify their tx.

See:
https://github.com/dapperlabs/dapper-flow-contracts/tree/c6496eeb26022b8ada7666e79f98d1fce5d9b158/dapper-storage-rent
for impl of storage check


satyamakgec:
@jrkhan let's hold this PR for a while.

cc @franklywatson

tpetrychyn:
stale closing

Issue #204, opened 2023-01-21
Provide a transaction to relink a topshot collection


bjartek:
#Why

Some users have topshot collections that are only linked with the TopShotCollectionPublic restrictions. Like this user 0x98daa38ecdf9972e

The current setup transaction https://github.com/dapperlabs/nba-smart-contracts/blob/master/transactions/user/setup_account.cdc sets up an account according to the NFT Metadata standard, but there is no transaction that I can see that provide a way for old users to relink their collection. 

# Backstory
A user in .find contacted us after somebody tried to send his dapper account a topshot moment. It failed to send so it was put into flowty lost and found. 

I have then try to remeed it from there for him, but that does not work since his collection is not linked according to the NFT standard. It only uses the TopshotCollectionPublic restriction and 3rd party tools have no way of knowing all of the different premutations. 

## How to fix
provide a relink transaction and surface that in the dapper wallet if a user has a collection that is linked wrong.

joshuahannan:
@jrkhan Is this something y'all can prioritize?

jrkhan:
Sure thing - will bring this up during backlog grooming today.

bjartek:
Gaia has this I belive. Talk to them.

PR #210, opened 2023-03-22
Use correct interface types in linking

joshuahannan:
The public links need to use all the types in `NonFungibleToken` and `TopShot` for these purposes so they are compatible with as many 3rd party apps as possible. The Top Shot team should make sure they're following this transaction's model for linking an account


jrkhan (approved):

jrkhan:
Out of curiosity, could this transaction also be used to relink account collections?
Per @bjartek request:
https://github.com/dapperlabs/nba-smart-contracts/issues/204

bjartek:
Not the way it is written right now no.

transactions/shardedCollection/setup_sharded_collection.cdc
```diff
@@ -25,7 +26,7 @@ transaction(numBuckets: UInt64) {
                 acct.unlink(/public/MomentCollection)
             }
 
-            acct.link<&{TopShot.MomentCollectionPublic}>(/public/MomentCollection, target: /storage/ShardedMomentCollection)
+            acct.link<&{NonFungibleToken.Receiver, NonFungibleToken.CollectionPublic, TopShot.MomentCollectionPublic}>(/public/MomentCollection, target: /storage/ShardedMomentCollection)
```
satyamakgec, line 29: Why not `MetadataViews.ResolverCollection` here ?

joshuahannan: the sharded collection doesn't implement that interface so we can't link it

PR #216, opened 2023-06-09
allow moment id casting to be either uint32 or uint64

zihehuang:
I observed the following casting error when processing MomentMinted events
```panic: interface conversion: cadence.Value is cadence.UInt32, not cadence.UInt64```


naguigui (approved):

rrrkren:
this was due to the encoding change that was already reverted on testnet. the uint32 was likely a different field.

PR #227, opened 2023-09-19
Fixing Sharded Collection Contract

javiersoto15:
Adding the borrowNFTSafe method to comply with the Collection Public interface and also to allow for the aggregator to be used


contracts/TopShotShardedCollection.cdc
```diff
@@ -35,8 +35,8 @@
 
 */
 
-import NonFungibleToken from 0xNFTADDRESS
-import TopShot from 0xTOPSHOTADDRESS
+import NonFungibleToken from 0x631e88ae7f1d7c20
```
sisyphusSmiling, line 38: Are these contract aliases intentionally replaced? Looks like other contracts in this repo have aliases

javiersoto15: That was my fault @sisyphusSmiling, i just fixed it

sisyphusSmiling (approved):

tpetrychyn (approved):

PR #229, opened 2023-12-10
Fast Break

HouseOfHufflepuff:
Fast Break is a game built on the NBA Top Shot contract. Players of Fast Break submit moments in their Top Shot collection to play the game. Each Fast Break contains n FastBreakStatistic the wallet must match or exceed.

When a player submits a lineup to Fast Break, a game token is minted to their Fast Break collection. This game token contains their submission data and links to the game. As NBA games are played, the metadata of the token (points & win) are updated. 

A Fast Break Run is a block of Fast Breaks. Winning a Fast Break puts you on the Fast Break Run leaderboard.

The resource FastBreakDaemon is reserved for the game oracle. The oracle is responsible for bringing trusted NBA data into its platform, computing points & wins, and updating the blockchain with the derived data. The FastBreakDaemon keeps the on-chain & off-chain games in sync.

Custodial submissions to Fast Break will be made by the game oracle in batch. Non-custodial submissions can be made by any wallet with Top Shots.

Screen Shot 2023-12-17 at 7 05 20 PM


joshuahannan (commented):
Can you add comments to all your types, fields, and functions in the contract? That would make the review process much easier for us. Thank you!

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,606 @@
+/*
+    Fast Break Game Contract
+    Author: Jeremy Ahrens jer.ahrens@dapperlabs.com
+*/
+
+import NonFungibleToken from 0xNFTADDRESS
+import TopShot from 0xTOPSHOTADDRESS
```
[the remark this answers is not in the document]
HouseOfHufflepuff: Yes, absolutely. I am doing this PR as part of a side project in my free time, but will prio it with @annatangzhao .

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,682 @@
...
+                serialNumber: self.serialNumber,
+                mintingDate: self.mintingDate,
+                topShots: self.topShots,
+                mintedTo: self.mintedTo
+            )
+        }
+
+        pub fun isWinner(): Bool {
+            let fastBreak : FastBreak.FastBreakGame = FastBreak.fastBreakGameByID[self.fastBreakGameID]!
+            let submission : FastBreak.FastBreakSubmission = fastBreak.submissions[self.mintedTo]!
```
turbolent, line 361: Here and elsewhere: No need for type annotations, simply use type inference: 
```suggestion
            let fastBreak = FastBreak.fastBreakGameByID[self.fastBreakGameID]!
            let submission = fastBreak.submissions[self.mintedTo]!
```

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,682 @@
...
+    pub let OraclePrivatePath:      PrivatePath
+
+    /// Contract variables
+    ///
+    pub var totalSupply:        UInt64
+
+    /// Metadata Dictionaries
+    ///
+    access(self) let fastBreakRunByID:        {String: FastBreakRun}
+    access(self) let fastBreakGameByID:           {String: FastBreakGame}
```
turbolent, line 77: No idea about the business logic, so may not apply/be possible: Maybe consider refactoring the architecture of how/where games are stored. Currently, the contract itself stores all games, which might become a scaling bottleneck. Is this really necessary?

Maybe try and see if account-based storage architecture could work here. For example, for a game with a "tournament"/"round", allow it to be created and stored in any account.

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,682 @@
...
+        ///
+        access(contract) fun updateStatus(status: String) { self.status = status }
+
+        /// Write a new win to the Fast Break Run leaderboard
+        ///
+        access(contract) fun incrementLeaderboardWins(wallet: Address) {
+            let leaderboard = self.leaderboard
+            var wins = leaderboard[wallet] ?? 0
+            wins = wins + 1
+            leaderboard[wallet] = wins
```
turbolent, line 120:
```suggestion
            leaderboard[wallet] = (leaderboard[wallet] ?? 0) + 1
```

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,682 @@
...
+    pub event FastBreakGameStatAdded(
+        fastBreakGameID: String,
+        name: String,
+        type: String,
+        valueNeeded: UInt64
+    )
+
+    /// Named Paths
+    ///
+    pub let CollectionStoragePath:  StoragePath
```
turbolent, line 65: Here and below: whitespace is off

turbolent (commented):
Great work!

Reviewed the code purely from a Cadence-code style perspective, I have no idea about the game logic, so cannot say if this works as expected

contracts/FastBreak.cdc
joshuahannan: agreed with this. It might be useful for users to store the games they participate in in their own account or something like that

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,682 @@
...
+        pub fun play(
+            fastBreakGameID: String,
+            topShots: [UInt64]
+        ): @FastBreak.NFT {
+            pre {
+                FastBreak.fastBreakGameByID.containsKey(fastBreakGameID): "no such fast break game"
+                FastBreak.fastBreakGameByID[fastBreakGameID]!.isPublic: "fast break game is private"
+            }
+
+            let acct = getAccount(self.owner?.address!)
```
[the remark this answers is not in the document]
joshuahannan: Not sure if this can really be a constant because the owner can change if the resource is moved to a different account and there wouldn't be a way to update it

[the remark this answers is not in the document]
turbolent: I see 👍 So depending on which variant is correct, either the initial owner keeps on being used, or always the current owner at each step of the function, maybe make it explicit that the code has that intent.

contracts/FastBreak.cdc
HouseOfHufflepuff: Yes, I agree and that's what I was going to do initially. It's more ideal for sure.

The problem that I could not figure out how to solve is that the game score for each user will be updated 100s of times by the oracle as the NBA games are going on. I can write to the token for custodial accounts I have the key for but don't think I can write to a non-custodial account token. Non-custodial accounts would not be able to get their token scored, afaik.

I don't write a ton of cadence so perhaps I am missing something, but wanted to explain why I took the direction of game object storage in the contract as opposed to wallets.

tpetrychyn: Also not knowing the inner-workings of Flow: does it change anything if the games and runs are resources owned by the admin wallet (instead of a struct)? Then when a tx needs to fetch details/scores about the game we would do a borrow operation on a specific game.

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,682 @@
...
+        ///
+        pub fun addStatToFastBreakGame(fastBreakGameID: String, name: String, type: String, valueNeeded: UInt64) {
+            let fastBreakGame: &FastBreak.FastBreakGame = (&FastBreak.fastBreakGameByID[fastBreakGameID] as &FastBreak.FastBreakGame?)!
+            let fastBreakStat : FastBreak.FastBreakStat = FastBreak.FastBreakStat(
+                name: name,
+                type: type,
+                valueNeeded: valueNeeded
+            )
+
+            fastBreakGame.addStat(stat: fastBreakStat)
```
tpetrychyn, line 614: Do we need to leave open the possibility of deleting or modifying a stat in the event of an incorrect stat getting added?

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,682 @@
...
+        /// Submit a Fast Break on behalf of a wallet
+        ///
+        pub fun submitFastBreak(wallet: Address, submission: FastBreak.FastBreakSubmission) {
+            let fastBreakGame: &FastBreak.FastBreakGame = (&FastBreak.fastBreakGameByID[submission.fastBreakGameID] as &FastBreak.FastBreakGame?)!
+            fastBreakGame.submitFastBreak(submission: submission)
+        }
+
+        /// Updates the submission scores of a Fast Break
+        ///
+        pub fun updateFastBreakScore(fastBreakGameID: String, wallet: Address, points: UInt64, win: Bool) {
```
tpetrychyn, line 641: Any considering needed for the new tie-break rule of lowest serial numbers? Or is that something that will be reconciled off-chain?

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,682 @@
...
+                self.fastBreakRunID = fastBreakRunID
+                self.stats = []
+                self.winner = nil
+            }
+        }
+
+        /// Get a wallet's active Fast Break Submission
+        ///
+        pub fun getFastBreakSubmissionByWallet(wallet: Address): FastBreak.FastBreakSubmission? {
+            let fastBreakSubmissions = self.submissions!
```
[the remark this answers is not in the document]
Deewai: can you explain why force-unwrap is bad?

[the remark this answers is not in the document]
joshuahannan: If the value that you're trying to unwrap is `nil`, then the force-unwrap will fail and the entire transaction execution will also fail with an unhelpful error message about the force unwrap failing, but no context. Force unwrap should really only be used when you are 100% sure that the optional will never be `nil`. 

Alternatively, you should use if-let to gracefully unwrap the optional and if it is `nil`, you can handle that without having to fail the entire transaction. This example, a getter function, should just return `nil` if `self.submissions` is `nil` to indicate that the submissions was `nil`. As a last resort, if you have to revert the transaction if the value is `nil`, you should just `panic()` with a descriptive error message, because that is much more helpful than just failing with none

contracts/FastBreak.cdc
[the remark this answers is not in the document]
joshuahannan: It shouldn't take much time, it is basically just a find and replace on all the contracts, transactions and scripts, then just updating the `contracts` and `templates` packages to use the new ones instead. I created an issue: https://github.com/dapperlabs/nba-smart-contracts/issues/230

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,682 @@
...
+/// Game & Oracle Contract for Fast Break
+///
+pub contract FastBreak: NonFungibleToken {
+
+    /// Contract events
+    ///
+    pub event ContractInitialized()
+    pub event Withdraw(id: UInt64, from: Address?)
+    pub event Deposit(id: UInt64, to: Address?)
+    pub event FastBreakRunCreated(
```
joshuahannan, line 18: I would recommend emitting the UUIDs for all the NFT in these events as well

contracts/FastBreak.cdc
joshuahannan: yeah, probably does make sense to just store them centrally.

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,682 @@
+/*
+    Fast Break Game Contract
+    Author: Jeremy Ahrens jer.ahrens@dapperlabs.com
+*/
+
+import NonFungibleToken from 0xNFTADDRESS
+import TopShot from 0xTOPSHOTADDRESS
+
+/// Game & Oracle Contract for Fast Break
```
joshuahannan, line 9: Valiant effort on the comments, but what I meant when I asked for comments was to have comments that actually give some context on what things are for. like for each event have a description of the specific context when the event is emitted, or for a type definition, what the type is for and how it is used in practice, and comments for the unclear fields

Additionally, some comments at the top of the file defining some common terminology, like "run", "game", "fatigue", etc

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,682 @@
...
+    /// Contract variables
+    ///
+    pub var totalSupply:        UInt64
+
+    /// Metadata Dictionaries
+    ///
+    access(self) let fastBreakRunByID:        {String: FastBreakRun}
+    access(self) let fastBreakGameByID:           {String: FastBreakGame}
+
+    /// A top-level Fast Break Run, the container for Fast Break Games
```
joshuahannan, line 79: What do you mean by "container for fast break games"

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,682 @@
...
+            }
+        }
+
+        /// Update status of the Fast Break Run
+        ///
+        access(contract) fun updateStatus(status: String) { self.status = status }
+
+        /// Write a new win to the Fast Break Run leaderboard
+        ///
+        access(contract) fun incrementLeaderboardWins(wallet: Address) {
```
joshuahannan, line 116: probably should use a different name than `wallet`, like `account`, or something. It might also be better to tie stats to resource IDs instead of accounts too because that is more accurate to the cadence programming model

contracts/FastBreak.cdc
joshuahannan, line 77: Any reason the IDs are strings and not integers?

contracts/FastBreak.cdc
joshuahannan: It also would allow people to trade and sell their fastbreak player records since the records are tied to NFTs instead of accounts, so it is another feature that takes advantage of web3 unique features

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,682 @@
...
+            topShots: [UInt64],
+        ) {
+            self.wallet = wallet
+            self.fastBreakGameID = fastBreakGameID
+            self.topShots = topShots
+            self.submittedAt = UInt64(getCurrentBlock().timestamp)
+            self.points = 0
+            self.win = false
+
+            emit FastBreakSubmissionSent(
```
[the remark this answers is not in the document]
joshuahannan: agreed, we shouldn't do any state changing operations in struct `init` blocks

joshuahannan (commented):
Please correct me if I am wrong, but it looks like the actual NFTs that you are defining here are just meant to only be used in a single game and are not used any time afterwards? If that is the case, it doesn't really make sense to make them the NFTs and it would make a lot more sense to make the NFT represent the player instead of the game, so it can be used indefinitely across multiple different games, can be moved to a new address if the player wants to for whatever reason, can be traded or sold on marketplaces, and can contain meaningful metadata about the player' history. 

In addition, you'll also need to have this contract implement `ViewResolver` and use metadata views for the NFTs and the contract

contracts/FastBreak.cdc
HouseOfHufflepuff: Will make the comment better, but a Fast Break occurs on any day there are NBA games. A Fast Break Run is a collection of Fast Breaks spanning about 2 weeks. We have rewards setup for hitting thresholds. 

Screen Shot 2023-12-21 at 12 18 59 PM

contracts/FastBreak.cdc
HouseOfHufflepuff: Artifact of doing the off-chain version first. I used strings there.

transactions/fastbreak/oracle/add_stat_to_game.cdc
```diff
@@ -0,0 +1,25 @@
...
+        self.oracleRef.addStatToFastBreakGame(
+            fastBreakGameID: fastBreakGameID,
+            name: name,
+            type: type,
+            valueNeeded: valueNeeded
+        )
+    }
+
+    post {
+        FastBreak.getFastBreakGameStats(id: fastBreakGameID).length != 0: "could not find fast break stat"
```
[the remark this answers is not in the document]
HouseOfHufflepuff: Good call. Removed this from tx in favor of a script & test.

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,682 @@
...
+                self.submissions = fb.submissions
+                self.fastBreakRunID = fb.fastBreakRunID
+                self.stats = fb.stats
+            } else {
+                self.id = id
+                self.name = name
+                self.isPublic = isPublic
+                self.submissionDeadline = submissionDeadline
+                self.numPlayers = numPlayers
+                self.status = "FAST_BREAK_OPEN"
```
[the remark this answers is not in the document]
HouseOfHufflepuff: Good call. Replaced with an enum.

transactions/fastbreak/user/play.cdc
```diff
@@ -0,0 +1,30 @@
...
+            ?? panic("could not borrow a reference to the collection receiver")
+    }
+
+    execute {
+
+        let nft <- self.gameRef.play(
+            fastBreakGameID: fastBreakGameID,
+            topShots: topShots
+        )
+        self.recipient.deposit(token: <- (nft as @NonFungibleToken.NFT))
```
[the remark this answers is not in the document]
HouseOfHufflepuff: I gave it a try and it looks like it is needed.

contracts/FastBreak.cdc
HouseOfHufflepuff: Resolving this convo as it appears a central ledger is operative here for the time being.

contracts/FastBreak.cdc
HouseOfHufflepuff: Currently this is a no in the off-chain game because it would strand wallets who already submitted to the fast break.

One change I want to make in the short term is to allow edits until the Fast Break is in an open state. But this may not ever get prioritized, so for the moment, you cannot edit stats once a fast break has been created. Must do so carefully like when we make sets & series.

lib/go/contracts/contracts.go
```diff
@@ -27,6 +27,7 @@ const (
 	defaultTopShotLockingAddress   = "TOPSHOTLOCKINGADDRESS"
 	defaultTopShotRoyaltyAddress   = "TOPSHOTROYALTYADDRESS"
 	defaultNetwork                 = "${NETWORK}"
+	fastBreakFile                  = "FastBreak.cdc"
```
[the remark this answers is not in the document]
HouseOfHufflepuff: Created an issue for this to be prioritized.

https://github.com/dapperlabs/nba-smart-contracts/issues/236

contracts/FastBreak.cdc
HouseOfHufflepuff: Am emitting this event when the game token is minted. Currently line 349. LMK if this is not what you meant.

emit FastBreakGameTokenMinted(
    id: self.id,
    fastBreakGameID: self.fastBreakGameID,
    serialNumber: self.serialNumber,
    mintingDate: self.mintingDate,
    topShots: self.topShots,
    mintedTo: self.mintedTo
)

contracts/FastBreak.cdc
HouseOfHufflepuff: I have renamed wallet to account. Good call.

I gave the concept of using a resourceId a lot of thought and it has advantages like you describe. I never considered that originally, so it was fun to think through. The reason I chose a flow account address here is two-fold. First, I made the off-chain version first and I need to tie non-custodial submissions to something I know about off-chain which is flowAddress. Next, this game token is "soul-bound" lite in that it records the address of who won it, but does not restrict ownership like a true soul-bound token would. There are a lot of opinions on this and have run this concept up to Dete for a related project and this is his preferred "trophy" implementation. Record the address of the winner and keep the token an NFT. People building on this can choose to use or discard this additional data of the token.

I do dig the concept of resourceIds here but our systems are built around flow address and am going to go this route to make integration easier.

contracts/FastBreak.cdc
HouseOfHufflepuff: Good call. I think that's a good future function because right now, the off-chain version does not allow this because if anyone submitted to the former stat, their record would get stranded.

To unblock that, I need to add a new state to the game like "DRAFT" and allow changes during this period while also restricting user submission when the Fast Break is in this state.

At this point, we don't support this off-chain so need to have the on-chain stats be immutable as well.

Perhaps a function to remove a Fast Break entirely on-mistake would be nice to have.

contracts/FastBreak.cdc
HouseOfHufflepuff: Absolutely, added depth to comments, especially field and game definitions.

contracts/FastBreak.cdc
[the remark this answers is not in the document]
HouseOfHufflepuff: Added context and intention into the comments of this function. Thanks.

HouseOfHufflepuff:
> Please correct me if I am wrong, but it looks like the actual NFTs that you are defining here are just meant to only be used in a single game and are not used any time afterwards? If that is the case, it doesn't really make sense to make them the NFTs and it would make a lot more sense to make the NFT represent the player instead of the game, so it can be used indefinitely across multiple different games, can be moved to a new address if the player wants to for whatever reason, can be traded or sold on marketplaces, and can contain meaningful metadata about the player' history.
> 
> In addition, you'll also need to have this contract implement `ViewResolver` and use metadata views for the NFTs and the contract

Yes, the intention is to create a game token that represents a single play of Fast Break. I chose this design because it maximizes the flexibility other contracts can have building on-top of Fast Break.

Having said that, I really like the idea of having a FastBreakPlayer resource or NFT. I think it adds a ton to the game. I started off creating it as a resource over here, but I am thinking it might be better as an NFT in its own contract. Would appreciate advice or collaboration on this PR on the FastBreakPlayer concept.

https://github.com/dapperlabs/nba-smart-contracts/pull/237

tpetrychyn (approved):
Reviewed via a slack huddle, LGTM as a v1

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,718 @@
...
+                FastBreak.fastBreakGameByID[fastBreakGameID]!.isPublic: "fast break game is private"
+            }
+
+            let acct = getAccount(self.owner?.address!)
+            let collectionRef = acct.getCapability(/public/MomentCollection)
+                                    .borrow<&{TopShot.MomentCollectionPublic}>()!
+
+            /// Must own Top Shots to play Fast Break
+            for flowId in topShots {
+                if !collectionRef.getIDs().contains(flowId) {
```
tpetrychyn, line 512: `getIDs()` gets all of the nft keys as an array from the collection, I wonder if there could be issue iterating over an array the length of the users collection `O(topShots*collectionSize)` times.

One suggestion is attempt to borrow each nft `collectionRef.borrowNFT(id: flowId)` which is O(1) map lookup

contracts/FastBreak.cdc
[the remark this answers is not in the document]
joshuahannan: @HouseOfHufflepuff reminder about this

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,708 @@
...
+
+            self.submissions[submission.accountAddress] = submission
+        }
+
+        /// Update the Fast Break score of an account
+        ///
+        access(contract) fun updateScore(accountAddress: Address, points: UInt64, win: Bool): Bool {
+            var isNewWin = false
+            let submissions = self.submissions
+            let submission: FastBreak.FastBreakSubmission = submissions[accountAddress]!
```
joshuahannan, line 252: Remember to use `if...let` instead of force unwrap

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,682 @@
...
+    }
+
+    /// Validate Fast Break Submission
+    ///
+    pub fun isValidSubmission(submissionDeadline: UInt64): Bool {
+        if submissionDeadline > UInt64(getCurrentBlock().timestamp) {
+            return true
+        }
+
+        return false
```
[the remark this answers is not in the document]
joshuahannan: @HouseOfHufflepuff reminder on this

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,682 @@
...
+    /// Get a Fast Break Game by Id
+    ///
+    pub fun getFastBreakGame(id: String): FastBreak.FastBreakGame? {
+        return FastBreak.fastBreakGameByID[id]
+    }
+
+    /// Get the game stats of a Fast Break
+    ///
+    pub fun getFastBreakGameStats(id: String): [FastBreak.FastBreakStat] {
+        let fastBreak = FastBreak.getFastBreakGame(id: id)!
```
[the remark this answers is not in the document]
joshuahannan: @HouseOfHufflepuff reminder on this. If any getter function can not access the information that is being queried, it shouldn't panic. It should either return `nil` or an empty version of the thing being queried. I would also take a look at all the other getter functions in this contract to make sure they all follow that pattern

joshuahannan (commented):
Just left a few reminders on some of bastians original review comments. Lets get those fixed now, and I'll be able to get to a full review soon

contracts/FastBreak.cdc
HouseOfHufflepuff: Good call out. I thought it was a map for some reason. Updated to your suggestion for the reasons you mentioned. Thank you.

contracts/FastBreak.cdc
HouseOfHufflepuff: I think I got all of these, thanks Josh.

joshuahannan (commented):
So I take it that you don't want to made the player a resource instead of tracking things by account address? It doesn't seem like you made any significant changes since the last time I looked at this

HouseOfHufflepuff:
@joshuahannan, I want to make the Player an NFT in a new contract and started working on it in the PR mentioned in my comment. Having said that, I do think a game token adds more to the goals of the game, is more experimental, and more composable, and I am planning to stick with this concept. However, I do like the idea of a player NFT as well, but correct me if I am wrong, that would need to be in a different contract. 

Either way, the core functionality I want in this contract is present in this PR. Have made all the suggestions Bastian made afaik. Thanks again for your time. 

https://github.com/dapperlabs/nba-smart-contracts/pull/229#issuecomment-1871637570

contracts/FastBreak.cdc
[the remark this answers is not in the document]
joshuahannan: Still need to address this @HouseOfHufflepuff

joshuahannan:
oh sorry, I think I missed that last part of your comment last week. I disagree that you are able to implement the player as a resource in a different contract. You're already tracking players by address in this contract, so if you were to try to graft a resource from a separate contract onto this as a player, it wouldn't work IMO because it can never be reconciled with the address tracking that you're doing here. It is a fundamental enough part of the game that you would need to track players by only their resource IDs and remove address tracking completely. That is also a more Cadence-best practices way to do it.

HouseOfHufflepuff:
Hey @joshuahannan, sorry for the delay as I am working on this over weekends. I implemented the Player resource and removed wallet data in favor of a resourceId. I like the concept a lot and feel like this really improved the contract. Thanks for pushing me to do this. Would appreciate your re-review.

javiersoto15 (approved):

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,802 @@
...
+            self.points = points
+            self.win = win
+        }
+
+    }
+
+    /// Resource for playing Fast Break
+    /// The Fast Break Player plays the game & mints game tokens
+    ///
+    pub resource Player: FastBreakPlayer {
```
joshuahannan, line 354:
```suggestion
    pub resource Player: FastBreakPlayer, NonFungibleToken.INFT {
```
I think it makes sense to make this an NFT also. No need to have a collection defined for it too, though, as long as it implements the interface, that should be fine, especially once Cadence 1.0 comes around

contracts/FastBreak.cdc
```diff
@@ -0,0 +1,802 @@
...
+        /// Each account must own a top shot collection to play fast break
+        ///
+        pub fun play(
+            fastBreakGameID: String,
+            topShots: [UInt64]
+        ): @FastBreak.NFT {
+            pre {
+                FastBreak.fastBreakGameByID.containsKey(fastBreakGameID): "no such fast break game"
+            }
+
```
joshuahannan, line 383: Here is where you could put the update to the mapping:
```suggestion
// updating player address mapping
if let ownerAddress = self.owner?.address {
    FastBreak.playerAccountMapping[self.id] = ownerAddress
}
```

joshuahannan (approved):
Looks good. I like the player resource model better. One thing you do need to be aware of with this model is that because most web3 people still care so much about addresses, people are probably still going to want to be able to see stats based on addresses instead of player IDs because they're stuck in the old ways and don't understand that tying things to resources is much better in terms of composability and extensibility. unfortunately, the'll want it either way.

There are a few ways you could handle this. You could just say no and tell people to adjust to this new model. That might work, but you would need to push this model pretty heavily in the design of the user facing stats and game. You could also emit addresses in events and such like we do with `Withdraw` and `Deposit` to at least have the events include an address. You just can get that from a player or collection resource by doing `self.owner?.address`.

Or, you could make all this information more accessible on-chain by having a mapping of playerIDs to addresses to show which address owns which player at any given time. You could keep this updated by just putting a line in one of the player methods that just keeps it updated automatically.

PR #256, opened 2024-08-08
Deprecate MomentCollectionPublic

fkenji:


fkenji:
Closing for now because we would need to refactor the tests first - they conflict due to referencing to the same resource path

PR #257, opened 2024-08-12
NBA-2865: Fixed event unit tests

judezhu:


lib/go/events/deposit.go
```diff
@@ -37,22 +36,11 @@ func (evt depositEvent) Owner() string {
 	return evt.To()
 }
 
-func (evt depositEvent) validate() error {
-	if evt["eventType"].(string) != EventDeposit {
```
fkenji, line 41: Is "eventType" not available as a field anymore?

fkenji (approved):

lib/go/events/deposit.go
judezhu: `eventType` is still available, but won't be accessible in this way. So we have to do update a lot of places to add this check. let me see if we can have a quicker way to check it.

PR #265, opened 2024-08-28
Added Pack Nft contract

judezhu:


judezhu:
see it here: https://github.com/dapperlabs/studio-platform-smart-contracts/blob/stable-cadence/pds/contracts/PackNFT_TopShot.cdc

PR #273, opened 2024-12-11
nba(jump ball): cadence

chumeston:
* Write initial contract.


contracts/JumpBall.cdc
```diff
@@ -48,109 +55,178 @@ access(all) contract JumpBall {
...
+                let previousOwner = self.ownership.remove(key: key)!
+                winnerCollection.deposit(token: <-nft)
+                emit NFTAwarded(gameID: self.id, nftID: key, previousOwner: previousOwner, winner: winner)
+            }
+        }
+    }
+
+    // Admin resource for game management
+    access(all) resource Admin {
+        access(all) fun determineWinner(gameID: UInt64, winnerCap: Capability<&{NonFungibleToken.Collection}>, stats: {UInt64: UInt64}) {
```
loic1, line 113: should receiver cap for both player be stored as part of the game so that the cap can be used here instead of being able to pass any cap as function arg?

contracts/JumpBall.cdc
```diff
@@ -48,109 +55,178 @@ access(all) contract JumpBall {
...
-        let gameID = self.nextGameID
-        self.games[gameID] <- create Game(id: gameID, creator: creator, opponent: opponent)
-        self.nextGameID = self.nextGameID + 1
-
-        // Update userGames mapping
-        self.addGameForUser(creator, gameID)
-        self.addGameForUser(opponent, gameID)
-
-        emit GameCreated(gameID: gameID, creator: creator)
+    access(all) fun createGame(creator: Address, startTime: UFix64, gameDuration: UFix64): UInt64 {
```
loic1, line 156: should validation be added for the address parameter, or is the intention that anyone can create a game with any address?

should a player resource be defined an the ability to create a game be scoped to it, so that games can be tied to each player resource's uuid?

contracts/JumpBall.cdc
```diff
@@ -48,109 +55,178 @@ access(all) contract JumpBall {
...
+        let gameID = JumpBall.nextGameID
+        JumpBall.games[gameID] <- create Game(id: gameID, creator: creator, startTime: startTime, gameDuration: gameDuration)
+        JumpBall.nextGameID = JumpBall.nextGameID + 1
+        JumpBall.addGameForUser(creator, gameID)
+        emit GameCreated(gameID: gameID, creator: creator, startTime: startTime)
         return gameID
     }
 
+    // Add an opponent to a game
+    access(all) fun addOpponent(gameID: UInt64, opponent: Address) {
```
loic1, line 166: same comment as above, do we need validation for the address param? or should this be part of a player resource?

contracts/JumpBall.cdc
```diff
@@ -0,0 +1,379 @@
...
+        }
+    }
+
+    // Helper function to get the current time (mocked in Cadence)
+    access(all) fun getCurrentTime(): UFix64 {
+        return UFix64(getCurrentBlock().timestamp)
+    }
+
+    // Securely call determine winner from the admin resource
+    access(all) fun determineWinner(gameID: String, stats: {UInt64: UInt64}) {
```
loic1, line 376: This can be called by anyone not just the admin account - this can be addressed by storing the Admin resource in the admin account rather that the contract itself and changing the modifier of the `determineWinner` Admin resource method to be called externally (in a tx) by the admin account (https://github.com/dapperlabs/nba-smart-contracts/pull/273#discussion_r1888895587)

contracts/JumpBall.cdc
```diff
@@ -0,0 +1,379 @@
...
+        access(all) fun setMetadata(key: String, value: AnyStruct) {
+            self.metadata[key] = value
+        }
+
+        access(all) fun getMetadata(key: String): AnyStruct? {
+            return self.metadata[key]
+        }
+
+        // Create a new game
+        access(all) fun createGame(gameID: String, startTime: UFix64, gameDuration: UFix64, selectedStatistic: String): String {
```
loic1, line 44: We should add an entitlement for `createGame` and `addOpponent` to ensure only the owner/entitled reference to the Player resource can call the methods in case a capability for the resource is published

contracts/JumpBall.cdc
```diff
@@ -0,0 +1,379 @@
...
+        access(all) fun getCreatorCap(): Capability<&{NonFungibleToken.Collection}> {
+            return self.creatorCap
+        }
+
+        // Getter for opponentCap
+        access(all) fun getOpponentCap(): Capability<&{NonFungibleToken.Collection}>? {
+            return self.opponentCap
+        }
+
+        access(all) fun setMetadata(key: String, value: AnyStruct) {
```
loic1, line 118: We should add an entitlement for `setMetadata` and other methods that should not be callable publicly after getting a reference to a Game resource using the `getGame` contract method

contracts/JumpBall.cdc
```diff
@@ -0,0 +1,379 @@
...
+
+        // Save the Player resource in storage
+        self.account.storage.save<@Player>(<- create Player(
+            address: self.account.address,
+            collectionCap: collectionCap
+        ), to: /storage/JumpBallPlayer)
+    }
+
+    // Helper functions for contract metadata
+    access(all) fun setMetadata(key: String, value: AnyStruct) {
```
loic1, line 294: This can be called by anyone - we can address by making it access(contract) and wrapping in admin resource, or we just don't have a setMetadata method if we are not currently making use of the metadata map

contracts/JumpBall.cdc
```diff
@@ -0,0 +1,379 @@
...
+        if JumpBall.getCurrentTime() < gameRef.startTime + gameRef.gameDuration {
+            panic("Game is still in progress.")
+        }
+
+        emit TimeoutClaimed(gameID: gameID, claimant: claimant)
+        gameRef.returnAllNFTs()
+    }
+
+    // Destroy a game and clean up resources
+    access(all) fun destroyGame(gameID: String) {
```
loic1, line 338: This can also be called publicly, maybe this should be part of the Player resource with logic to validate the creator/opponent address? And/or this could be part of the Admin resource?

contracts/JumpBall.cdc
```diff
@@ -0,0 +1,379 @@
...
+        }
+
+        access(contract) fun awardWinner(game: &Game, winner: Address, winnerCap: Capability<&{NonFungibleToken.Collection}>) {
+            emit WinnerDetermined(gameID: game.id, winner: winner)
+            game.transferAllToWinner(winner: winner, winnerCap: winnerCap)
+        }
+    }
+
+    // Factory function to create a new Player resource
+    access(all) fun createPlayer(playerAddress: Address, collectionCap: Capability<&{NonFungibleToken.Collection}>): @Player {
```
loic1, line 260: Any `playerAddress` and `collectionCap` can be passed here not necessarily those of the account calling this method

We might be able to address this:
- By using Player resource UUIDs rather than addresses to tie games to users so that addresses are not needed anymore, or
- By using an entitled capability param here and getting the address from the capability itself, though we should avoid storing the capability if it's not useful (it looks like `collectionCap` is not actually used?). Also the address value would remain the same in case the player resource is transferred to someone else and there may be limitations related to that

contracts/JumpBall.cdc
```diff
@@ -0,0 +1,379 @@
...
+        // Add an opponent to a game
+        access(all) fun addOpponent(gameID: String) {
+            let gameRef = &JumpBall.games[gameID] as &Game?
+                    ?? panic("Game does not exist.")
+
+            if !self.collectionCap.check() {
+                panic("Player does not have a valid collection capability to join a game.")
+            }
+
+            gameRef.setOpponent(opponentAddress: self.address, collectionCap: self.collectionCap)
```
loic1, line 73: It looks like the game creator address is added instead of opponent address? Or maybe I'm misunderstanding something about the expected flow/caller for `addOpponent`?

contracts/JumpBall.cdc
```diff
@@ -0,0 +1,379 @@
...
+    access(all) event NFTDeposited(gameID: String, nftID: UInt64, owner: Address)
+    access(all) event NFTReturned(gameID: String, nftID: UInt64, owner: Address)
+    access(all) event NFTAwarded(gameID: String, nftID: UInt64, previousOwner: Address, winner: Address)
+    access(all) event WinnerDetermined(gameID: String, winner: Address)
+    access(all) event TimeoutClaimed(gameID: String, claimant: Address)
+
+    // Player resource for game participation
+    access(all) resource Player {
+        access(all) let address: Address
+        access(all) let collectionCap: Capability<&{NonFungibleToken.Collection}>
```
Deewai, line 25: this is for sending the nft to the user if they win?

Deewai (commented):
i think some things are missing, like how player resource interfaces with the game, will be nice to have some tests with transactions and scripts that define how the contract is to be used.
Also it's not clear if this is supposed to support 1 nft type or it's meant to be generic. Might also be better to not save capabilities but instead get the capability for that nft collection using metadataviews to resolve the collection public path

PR #276, opened 2025-01-09
nba(fb): cadence

chumeston:
* Decrease computation from local storing.


loic1 (approved):

loic1:
Quick fix already added in https://github.com/dapperlabs/nba-smart-contracts/pull/278

PR #280, opened 2025-02-07
Add creator token functionality & interfaces

loic1:
Add on-chain royalties functionality and creator token interface conformances

For reference:
- https://github.com/limitbreakinc/creator-token-standards
- https://docs.opensea.io/docs/creator-fee-enforcement

### DIFF
--- a/evm-bridging/script/Deploy.s.sol
+++ b/evm-bridging/script/Deploy.s.sol
@@ -2,6 +2,7 @@
 pragma solidity 0.8.24;
 
 import {Script} from "forge-std/Script.sol";
+import "forge-std/console.sol";
 import {Upgrades} from "openzeppelin-foundry-upgrades/src/Upgrades.sol";
 import {BridgedTopShotMoments} from "../src/BridgedTopShotMoments.sol";
 
@@ -13,13 +14,13 @@ contract DeployScript is Script {
         vm.startBroadcast(vm.envUint("DEPLOYER_PRIVATE_KEY"));
         console.log("Deployer address:", msg.sender);
 
-        // Set contract initialization parameters
+        // Set testnet contract initialization parameters
         address owner = msg.sender;
-        string memory name = "Bidged NBA TopShot Moments";
+        string memory name = "NBA Top Shot";
         string memory symbol = "TOPSHOT";
-        string memory cadenceNFTAddress = "cadenceNFTAddress";
-        string memory cadenceNFTIdentifier = "cadenceNFTIdentifier";
-        string memory contractURI = "contractURI";
+        string memory cadenceNFTAddress = "877931736ee77cff";
+        string memory cadenceNFTIdentifier = "A.877931736ee77cff.NFT";
+        string memory contractURI = "add-contract-URI-here";
 
         // Deploy NFT contract using UUPS proxy for upgradeability
         address proxyAddr = Upgrades.deployUUPSProxy(
--- a/evm-bridging/src/BridgedTopShotMoments.sol
+++ b/evm-bridging/src/BridgedTopShotMoments.sol
@@ -11,6 +11,7 @@ import {Initializable} from "@openzeppelin/contracts-upgradeable/proxy/utils/Ini
 import {IERC721Metadata} from "@openzeppelin/contracts/token/ERC721/extensions/IERC721Metadata.sol";
 import {IERC721Enumerable} from "@openzeppelin/contracts/token/ERC721/extensions/IERC721Enumerable.sol";
 import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
+import {IERC2981} from "@openzeppelin/contracts/interfaces/IERC2981.sol";
 import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
 
 import {ICreatorToken, ILegacyCreatorToken} from "./interfaces/ICreatorToken.sol";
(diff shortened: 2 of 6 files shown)


Deewai (approved):

evm-bridging/script/Deploy.s.sol
```diff
@@ -13,13 +14,13 @@ contract DeployScript is Script {
...
+        // Set testnet contract initialization parameters
         address owner = msg.sender;
-        string memory name = "Bidged NBA TopShot Moments";
+        string memory name = "NBA Top Shot";
         string memory symbol = "TOPSHOT";
-        string memory cadenceNFTAddress = "cadenceNFTAddress";
-        string memory cadenceNFTIdentifier = "cadenceNFTIdentifier";
-        string memory contractURI = "contractURI";
+        string memory cadenceNFTAddress = "877931736ee77cff";
+        string memory cadenceNFTIdentifier = "A.877931736ee77cff.NFT";
```
vdojnov, line 22: I think this should be `A. 877931736ee77cff.TopShot`? 🤔

evm-bridging/src/BridgedTopShotMoments.sol
```diff
@@ -110,4 +125,79 @@ contract BridgedTopShotMoments is
...
+        return _royaltyInfo.royaltyAddress;
+    }
+
+    function royaltyBasisPoints() external view returns (uint256) {
+        return _royaltyInfo.royaltyBps;
+    }
+
+    function royaltyInfo(
+        uint256,
+        /* _tokenId */
```
vdojnov, line 157: Curious why we do this? Is it cuz there is an interface that requires 2 params?

evm-bridging/src/BridgedTopShotMoments.sol
```diff
@@ -110,4 +125,79 @@ contract BridgedTopShotMoments is
...
+    }
+
+    function _beforeTokenTransfers(
+        address from,
+        address to,
+        uint256 startTokenId,
+        uint256 /* quantity */
+    ) internal virtual {
+        if (from != address(0) && to != address(0)) {
+            // Call the transfer validator if one is set.
```
vdojnov, line 191: If one is set, or if both are set? `from != address(0) && to != address(0)` vs `from != address(0) || to != address(0)`

evm-bridging/script/Deploy.s.sol
vdojnov: Also should we some how parameterize this, so we don't accidentally deploy the testnet address to mainnet

loic1: Good call, should be `A.877931736ee77cff.TopShot.NFT`

It looks like we need to create a separate deploy script in go using overflow library to make flow-cli calls and be able to sign with kms, will parameterize there and keep this deploy script if needed for testing

judezhu (approved):
Looks good to me. Probably need a bit more time to read and understand how the enforcement of the royalty works. And take a look what APIs are actually exposed. We can sync this later if you wanna go ahead of the current implementation.

evm-bridging/src/BridgedTopShotMoments.sol
loic1: Yea exactly this is to for `IERC2981` interface (suggested royalties), I don’t think it’s needed for now but the royalty config could vary depending in token id

evm-bridging/src/BridgedTopShotMoments.sol
loic1: We use this condition so that the hook excludes minting and burning (only applies to regular transfers)

evm-bridging/src/BridgedTopShotMoments.sol
```diff
@@ -104,10 +120,90 @@ contract BridgedTopShotMoments is
         override(ERC721Upgradeable, ERC721EnumerableUpgradeable)
         returns (address)
     {
+        // Add the beforeTokenTransfer hook
+        _beforeTokenTransfer(_ownerOf(tokenId), to, tokenId);
```
loic1, line 124: `_beforeTokenTransfer` hook was removed from the v5 of OpenZeppelin libraries so we need to add it here in the `_update` method, which is called when transferring, minting, and burning tokens

judezhu (approved):
LGTM

evm-bridging/src/BridgedTopShotMoments.sol
```diff
@@ -104,10 +120,90 @@ contract BridgedTopShotMoments is
...
+        return _royaltyInfo.royaltyBps;
+    }
+
+    /**
+     * @dev Implements the IERC2981 interface.
+     */
+    function royaltyInfo(
+        uint256 /* _tokenId */,
+        uint256 _salePrice
+    ) external view returns (address receiver, uint256 royaltyAmount) {
```
sisyphusSmiling, line 166: Does this mean the contract needs to implement the `IERC2981` interface?

sisyphusSmiling (approved):
Approved! Left a couple minor comments

evm-bridging/src/BridgedTopShotMoments.sol
```diff
@@ -23,12 +28,22 @@ contract BridgedTopShotMoments is
...
 {
     string public cadenceNFTAddress;
     string public cadenceNFTIdentifier;
     string public contractMetadata;
     string private _customSymbol;
+    RoyaltyInfo private _royaltyInfo;
+
+    error InvalidRoyaltyBasisPoints(uint256 basisPoints);
+    error RoyaltyAddressCannotBeZeroAddress();
+    event RoyaltyInfoUpdated(address receiver, uint256 bps);
```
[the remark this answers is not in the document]
loic1: Good call, thanks! Just checked and [seadrop](https://github.com/ProjectOpenSea/seadrop/blob/main/src/interfaces/ISeaDropTokenContractMetadata.sol#L66) contracts deployed from opensea are not using the keyword so I think we can keep as is, also this event should be emitted very infrequently

evm-bridging/src/BridgedTopShotMoments.sol
loic1: Added inheritance, thank you! (`IERC2981` also returned in supportsInterface)

zhangchiqing (approved):

PR #282, opened 2025-02-17
Add flow deploy script

loic1:
Add script deploying evm contracts from flow in order to sign transactions with account key details specified in `flow.json` / kms signers - this replaces the Solidity deploy script used for initial testing


evm-bridging/foundry.toml
```diff
@@ -1,7 +1,7 @@
 [profile.default]
 solc_version = "0.8.24"
 optimizer = true
-optimizer_runs = 200
+optimizer_runs = 50000
```
loic1, line 4: @judezhu increased optimizer runs to make function executions less expensive

More details: 

> Before you deploy your contract, activate the optimizer when compiling using solc --optimize --bin sourceFile.sol. By default, the optimizer will optimize the contract assuming it is called 200 times across its lifetime (more specifically, it assumes each opcode is executed around 200 times). If you want the initial contract deployment to be cheaper and the later function executions to be more expensive, set it to --optimize-runs=1. If you expect many transactions and do not care for higher deployment cost and output size, set --optimize-runs to a high number. This parameter has effects on the following (this might change in the future):
> 
>  - the size of the binary search in the function dispatch routine
> 
>  - the way constants like large numbers or strings are stored

https://docs.soliditylang.org/en/latest/using-the-compiler.html#optimizer-options

judezhu (approved):
nice!

PR #287, opened 2025-02-21
Add extra deployment details

loic1:


evm-bridging/main.go
```diff
@@ -68,11 +68,10 @@ var configByNetwork = map[string]config{
 		flowEvmBridgeCoaAddr:            "0x00000000000000000000000249250a5c27ecab3b",
 		bridgeDeployedTopshotERC721Addr: "0x50AB3a827aD268e9D5A24D340108FAD5C25dAD5f",
 		transferValidatorAddr:           "0xA000027A9B2802E1ddf7000061001e5c005A0000", // StrictAuthorizedTransferSecurityRegistry
-		// TODO: get royalty recipient
-		royaltyRecipientAddr: placeholderEvmAddress,
-		rpcUrl:               "https://mainnet.evm.nodes.onflow.org",
-		verifierUrl:          "https://evm.flowscan.io/api",
-		verifierProvider:     "blockscout",
+		royaltyRecipientAddr:            "0x000000000000000000000002b4c7594cb1b9e654", // owned by 0x065fb66f8d2d780a
+		rpcUrl:                          "https://mainnet.evm.nodes.onflow.org",
```
judezhu, line 72: should we add comment which key has access to this account?

judezhu (approved):
nice!

evm-bridging/main.go
loic1: signer details are in internal doc and we can get the public key with flow query or [flowview](https://www.flowview.app/account/0x065fb66f8d2d780a/key)

PR #288, opened 2025-02-21
Add EVMBridgedMetadata view

loic1:
Changes:
- Add `EVMBridgedMetadata` in TopShot contract to optimize gas consumption when bridging to Flow EVM since metadata is fetched via API in EVM and serialization from Cadence isn't needed


contracts/TopShot.cdc
```diff
@@ -801,6 +802,17 @@ access(all) contract TopShot: NonFungibleToken {
                     return TopShot.resolveContractView(resourceType: nil, viewType: Type<MetadataViews.NFTCollectionDisplay>())
                 case Type<CrossVMMetadataViews.EVMPointer>():
                     return TopShot.resolveContractView(resourceType: nil, viewType: Type<CrossVMMetadataViews.EVMPointer>())
+                case Type<MetadataViews.EVMBridgedMetadata>():
+                    // Project-defined ERC721 EVM contract stores baseURI, name, and symbol in its own contract storage
+                    // Name, symbol, and baseURI below are only used for legacy bridge-deployed ERC721 contract
+                    return MetadataViews.EVMBridgedMetadata(
+                        name: "NBA-Top-Shot",
+                        symbol: "NBAT",
+                        uri: MetadataViews.URI(
```
loic1, line 811: `baseURI` should be set to `https://metadata-api.production.studio-platform.dapperlabs.com/v1/topshot/moment/` on mainnet and left empty on testnet since the staging metadata api isn't exposed publicly

contracts/TopShot.cdc
```diff
@@ -801,6 +802,17 @@ access(all) contract TopShot: NonFungibleToken {
                     return TopShot.resolveContractView(resourceType: nil, viewType: Type<MetadataViews.NFTCollectionDisplay>())
                 case Type<CrossVMMetadataViews.EVMPointer>():
                     return TopShot.resolveContractView(resourceType: nil, viewType: Type<CrossVMMetadataViews.EVMPointer>())
+                case Type<MetadataViews.EVMBridgedMetadata>():
+                    // Project-defined ERC721 EVM contract stores baseURI, name, and symbol in its own contract storage
+                    // Name, symbol, and baseURI below are only used for legacy bridge-deployed ERC721 contract
+                    return MetadataViews.EVMBridgedMetadata(
+                        name: "NBA-Top-Shot",
+                        symbol: "NBAT",
```
vdojnov, line 810:
```suggestion
                        symbol: "NBATS",
```

loic1: `NBAT` is the name the bridge already gave by default to the legacy erc721 contract (see [flowscan](https://evm-testnet.flowscan.io/token/0xB3627E6f7F1cC981217f789D7737B1f3a93EC519)), I don't think we need to change it and it doesn't matter for the new contract, which maintains its own name/symbol

loic1: We won't need to pass anything after the new contract is onboarded to the bridge, which will deactivate bridging to the legacy contract - so at that point, we can replace these with empty string or achieve similar outcome with the upcoming `EVMBytesMetadata` new view

judezhu (approved):
nice!

PR #302, opened 2026-06-01
refactor: restrict access(all) var fields to access(contract) per Cadence best practices

fkenji:
## Summary

An external security researcher reported that several `access(all) var` fields across our contracts could allow unauthenticated writes in Cadence 1.0. After investigation and live testing on mainnet, we confirmed that **direct external assignment to `access(all) var` fields is blocked at runtime** — the Cadence runtime rejects such transactions with an explicit access control error:

```
error: cannot assign to `totalSupply`: field has `all` access
fields with `all` access cannot be directly assigned to
```

So there is **no actively exploitable vulnerability** for primitive-typed fields (`UInt64`, `UInt32`, `UFix64`, `String`). This is consistent with the Cadence security docs and matches Cadence 0.x behaviour for `pub var`.

However, the access modifiers were still incorrect by best-practice standards, and one real concern remains for **mutable container fields** (arrays and dictionaries): `access(all)` blocks reassignment of the field itself, but mutations *inside* the container (`.append()`, dictionary writes) through a reference may still be possible. This patch addresses both.

This PR is a **best-practices / hygiene fix**, not a critical security patch.

## What is actually at risk

| Field type | Direct assignment blocked? | Container mutation risk? |
|---|---|---|
| Primitive (`UInt64`, `UFix64`, `String`) | ✅ Yes — runtime rejects | N/A |
| Array (`[T]`) | ✅ Yes — reassignment blocked | ⚠️ Append/remove through reference may be possible |
| Dictionary (`{K: V}`) | ✅ Yes — reassignment blocked | ⚠️ Key mutation through reference may be possible |

The **marketplace capability type** (`&SaleCollection` vs `&{Market.SalePublic}`) is a separate real concern — the issued capability was broader than necessary regardless of field writability, violating principle of least privilege.

## Changes

- **`contracts/TopShot.cdc`** — Changed 4 contract-level fields to `access(contract) var`; added 4 public `view` getter functions (`getCurrentSeries`, `getNextPlayID`, `getNextSetID`, `getTotalSupply`)
- **`contracts/TopShotIPFSResolver.cdc`** — Changed `gateway` to `access(contract) var`; added `getGateway()` view getter
- **`contracts/TopShotMarketV3.cdc`** — Changed `SaleCollection.cutPercentage` to `access(self) var`; added `getCutPercentage()` view getter and an `access(TopShotMarketV3.Update)` entitlement-gated `changeCutPercentage()` setter
- **`contracts/FastBreakV1.cdc`** — Changed `totalSupply` and `nextPlayerId` contract fields to `access(contract) var` with public getters; restricted mutable struct fields on `FastBreakRun`, `FastBreakGame`, and `FastBreakSubmission` to `access(contract) var`
- **`transactions/marketV3/create_sale.cdc`** and **`create_start_sale.cdc`** — Narrowed issued capability type from `&TopShotMarketV3.SaleCollection` to `&{Market.SalePublic}` (genuine least-privilege fix)
- **Call-site updates** — 9 scripts/transactions updated to use the new getter functions

## Testing

- [ ] Confirm all existing scripts that read these fields still work correctly via the new getter functions
- [ ] Confirm `SaleCollection.changeCutPercentage()` succeeds with the `Update` entitlement and fails without it
- [ ] Run the full existing test suite to confirm no regressions

## Notes

- `Market.cdc` also contains `access(all) var cutPercentage` — out of scope for this patch, should be followed up separately.
- Reported by an external security researcher (Bertie). Confirmed non-exploitable after live mainnet testing by the Flow team.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


transactions/marketV3/create_sale.cdc
```diff
@@ -29,7 +29,7 @@ transaction(tokenReceiverPath: PublicPath, beneficiaryAccount: Address, cutPerce
         acct.storage.save(<-collection, to: TopShotMarketV3.marketStoragePath)
 
         acct.capabilities.publish(
-            acct.capabilities.storage.issue<&TopShotMarketV3.SaleCollection>(TopShotMarketV3.marketStoragePath),
+            acct.capabilities.storage.issue<&{Market.SalePublic}>(TopShotMarketV3.marketStoragePath),
```
Deewai: it should be fine to issue a cap for the concrete type, users will need the correct entitlement to access non public methods

transactions/marketV3/create_start_sale.cdc
```diff
@@ -70,7 +70,7 @@ transaction(tokenReceiverPath: PublicPath, beneficiaryAccount: Address, cutPerce
         
             // create a public link to the sale collection
            acct.capabilities.publish(
-                acct.capabilities.storage.issue<&TopShotMarketV3.SaleCollection>(TopShotMarketV3.marketStoragePath),
+                acct.capabilities.storage.issue<&{Market.SalePublic}>(TopShotMarketV3.marketStoragePath),
```
Deewai: same as above

Deewai (approved):

